PR #1, opened 2023-05-29
Performance improvements & speedtest clients

jagerman:
This PR adds a speedtest client/server to diagnose performance and, based on that, includes significant performance improvements in how we send/receive packets.

See individual commits (especially `Packet batching, sendmmsg, GSO, performance`) for the details.

## Known issue

Network close is broken by this PR for two reasons:
- During Network::close() we do a walk to close handles, but uvw walk assumes it owns *all* libuv types with data elements, and so it blindly casts our uv_udp_t to a uvw::udp_handle, joy!
- Our uv_udp_t destructors *should* be closing on destruction, but aren't (I think because our shared_ptr udp's are shared in too many places and so the destructors aren't firing).

These are not going to be fixed here; we will instead refactor endpoint to de-tangle the socket ownership and management.


include/quic/connection.hpp
```diff
@@ -24,7 +24,7 @@ namespace oxen::quic
     class Client;
     class Handler;
 
-    class Connection : public std::enable_shared_from_this<Connection>
```
dr7ana: @jagerman what's the motivation behind removing this? i was thinking it may be useful at some point to hold a weak pointer to the connection in `Stream.user_data`

include/quic/connection.hpp
```diff
@@ -92,9 +92,15 @@ namespace oxen::quic
...
 
+        /// Returns a pointer to the owning Server, if this is a Server connection, nullptr
+        /// otherwise.
         Server* server();
+        const Server* server() const;
 
+        /// Returns a pointer to the owning Client, if this is a Client connection, nullptr
+        /// otherwise.
         Client* client();
+        const Client* client() const;
```
dr7ana, line 103: would it be better to do an operator overload for the client and server classes s.t. we can pass them as const?

include/quic/connection.hpp
jagerman: No reason we can't re-add it later if we want it, but for now it isn't in use.

include/quic/connection.hpp
```diff
@@ -109,11 +109,9 @@ namespace oxen::quic
         Client* client();
         const Client* client() const;
 
-        void schedule_retransmit(uint64_t ts = 0);
-
-        int init_gnutls(Client& client);
+        /// Returns a pointer to the owning endpoint; identity can be
```
dr7ana: was this comment a leftover from resolving a merge conflict, or did the function itself get dropped? i dont see anything in `connection.cpp`

jagerman: Looks like it came in the server stream creation code.  I'll remove it.

tewinget (approved):
LGTM

PR #2, opened 2023-05-31
Refactored stream creation

dr7ana:
- Stream is not created when the first call to extend_max_streams_bidi is made at client-side handshake completion
- Streams can be created and data can be pushed down whether the connection is ready or not
- When ready, streams are moved from pending_streams to ready map, and broadcast as usual
- Unit test 004 testing the above

### DIFF
--- a/include/quic/client.hpp
+++ b/include/quic/client.hpp
@@ -32,7 +32,7 @@ namespace oxen::quic
 
         ~Client();
 
-        const std::shared_ptr<Stream>& open_stream(
+        std::shared_ptr<Stream> open_stream(
                 stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr);
 
         std::shared_ptr<uvw::UDPHandle> get_handle(Address& addr) override;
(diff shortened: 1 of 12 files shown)


include/quic/client.hpp
```diff
@@ -32,7 +32,7 @@ namespace oxen::quic
 
         ~Client();
 
-        const std::shared_ptr<Stream>& open_stream(
+        const std::shared_ptr<Stream> open_stream(
```
jagerman:
```suggestion
        std::shared_ptr<Stream> open_stream(
```

include/quic/connection.hpp
```diff
@@ -78,11 +78,14 @@ namespace oxen::quic
...
-        std::function<void(Connection&)> on_stream_available;
         std::function<void(Connection&)> on_closing;  // clear immediately after use
 
-        const std::shared_ptr<Stream>& open_stream(
-                stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr);
+        // change to check_pending_streams, do not create after while loop
+        void check_pending_streams(
+            int available, stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr);
+
+        const std::shared_ptr<Stream> get_new_stream(
```
jagerman:
```suggestion
        std::shared_ptr<Stream> get_new_stream(
```

jagerman (approved):

PR #3, opened 2023-06-01
Exposed callback for stream open

dr7ana:
- separate function than stream data callback
- only called on successful stream open (when opened by remote endpoint)

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -91,6 +91,12 @@ namespace oxen::quic
         stream_data_cb = std::move(func);
     }
 
+    void ServerContext::handle_serverctx_opt(stream_open_callback_t& func)
+    {
+        log::trace(log_cat, "Server given data callback");
+        stream_open_cb = std::move(func);
+    }
+
     ServerContext::~ServerContext()
     {
         for (auto& h : udp_handles)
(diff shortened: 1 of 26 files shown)


src/context.cpp
```diff
@@ -91,6 +91,12 @@ namespace oxen::quic
         stream_data_cb = std::move(func);
     }
 
+    void ServerContext::handle_serverctx_opt(stream_open_callback_t& func)
```
jagerman: This should be taken by value because, if it's a non-const lvalue reference, then you could end up stealing my value away from me.  For instance if I did this:
```C++
std::function<uint64_t(Stream&)> open_callback = [](Stream&) { /*... */; };

std::list<Server> servers;
for (int i = 0; i < 3; i++) {
    servers.emplace(open_callback);
}
```
then the *first* Server construction would end up stealing my function and leaving my `open_callback` empty, so that an empty function gets passed into the next ones.  (The same goes for the other callbacks, though I changed those in #1).

dr7ana: > (The same goes for the other callbacks, though I changed those in https://github.com/oxen-io/oxen-libquic/pull/1)

Completely agree, i remember speaking about this earlier today (that's also why I did it the same, less optimal, way). I figured it might be easier to leave them the same, in case it gave you a weird merge conflict where you changed the rest of them. I can change them all on this PR too if you prefer 👍🏼

jagerman (approved):

PR #4, opened 2023-06-02
Thread safety

tewinget:
Adds an event loop "caller" concept to the code, such that things which must run on the event loop thread call that and run either now (if already on that thread) or are queued to run on the next event loop tick.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -9,6 +9,7 @@ extern "C"
 #include <cstdio>
 #include <cstring>
 #include <stdexcept>
+#include <future>
 
 #include "connection.hpp"
 #include "context.hpp"
@@ -59,11 +60,25 @@ namespace oxen::quic
     std::shared_ptr<Stream> Client::open_stream(stream_data_callback_t data_cb, stream_close_callback_t close_cb)
     {
         log::trace(log_cat, "Opening client stream...");
-        auto ctx = reinterpret_cast<ClientContext*>(context.get());
 
-        auto conn = get_conn(ctx->conn_id);
+        std::promise<std::shared_ptr<Stream>> p;
+        auto f = p.get_future();
+        handler->call([&data_cb, &close_cb, &p, this](){
+            try
+            {
+                auto ctx = reinterpret_cast<ClientContext*>(context.get());
 
-        return conn->get_new_stream(std::move(data_cb), std::move(close_cb));
+                auto conn = get_conn(ctx->conn_id);
+
+                p.set_value(conn->get_new_stream(std::move(data_cb), std::move(close_cb)));
+            }
+            catch (...)
+            {
+                p.set_exception(std::current_exception());
+            }
+        });
+
+        return f.get();
     }
 
     std::shared_ptr<uvw::udp_handle> Client::get_handle(Address& addr)
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -66,9 +66,7 @@ namespace oxen::quic
         handler->call([&data_cb, &close_cb, &p, this]() {
             try
             {
-                auto ctx = reinterpret_cast<ClientContext*>(context.get());
-
-                auto conn = get_conn(ctx->conn_id);
+                auto conn = get_conn(context->conn_id);
 
                 p.set_value(conn->get_new_stream(std::move(data_cb), std::move(close_cb)));
             }
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -8,8 +8,8 @@ extern "C"
 
 #include <cstdio>
 #include <cstring>
-#include <stdexcept>
 #include <future>
+#include <stdexcept>
 
 #include "connection.hpp"
 #include "context.hpp"
@@ -63,7 +63,7 @@ namespace oxen::quic
 
         std::promise<std::shared_ptr<Stream>> p;
         auto f = p.get_future();
-        handler->call([&data_cb, &close_cb, &p, this](){
+        handler->call([&data_cb, &close_cb, &p, this]() {
             try
             {
                 auto ctx = reinterpret_cast<ClientContext*>(context.get());
--- a/src/network.cpp
+++ b/src/network.cpp
@@ -30,6 +30,7 @@ namespace oxen::quic
         signal_config(); // TODO: probably remove this, as the library user should handle signals
 
         loop_thread = std::make_unique<std::thread>([this](){ ev_loop->run(); });
+        running.store(true);
         quic_manager = std::make_shared<Handler>(ev_loop, loop_thread->get_id(), *this);
     }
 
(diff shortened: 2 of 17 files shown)


src/network.cpp
```diff
@@ -64,13 +66,59 @@ namespace oxen::quic
...
+                    auto s = handle->sock();
+                    log::debug(log_cat, "closing client UDPHandle bound to {}:{}", s.ip, s.port);
+                    handle->close();
+                }
+                for (auto& [addr, handle] : mapped_server_addrs)
+                {
+                    auto s = handle->sock();
+                    log::debug(log_cat, "closing server UDPHandle bound to {}:{}", s.ip, s.port);
+                    handle->close();
+                }
```
dr7ana: Nit-pick: I know that if `mapped_{client,server}_addrs` is empty, it will just exit out of the for-loop without iterating through it, but for some reason not explicitly checking sticks out to me in a bad way

Nittier-pick: it could probably be `const auto&`, since calling `handle->close()` doesn't modify the underlying object

src/network.cpp
```diff
@@ -64,13 +66,59 @@ namespace oxen::quic
...
+                if (loop_thread)
+                {
+                    ev_loop->walk(uvw::Overloaded{
+                            [](uvw::UDPHandle&& h) {
+                                auto s = h.sock();
+                                log::debug(log_cat, "closing UDPHandle bound to {}:{}", s.ip, s.port);
+                                h.close();
+                            },
+                            [](auto&&) {}});
```
dr7ana: We're going to have to had overloads here for the other types of handles (async, timer, etc). 

The example uvw has in their documentation of closing all handles through a generic callback doesn't work.
[Jason](https://github.com/skypjack/uvw/issues/239) brought up this issue 2 years ago, so it's great it's still present. Hopefully this would be fixed when we bump or replace uvw/libuv

src/stream.cpp
```diff
@@ -69,22 +69,24 @@ namespace oxen::quic
...
-        else
-        {
-            is_closing = is_shutdown = true;
-            log::info(log_cat, "Closing stream (ID: {}) with error code {}", stream_id, ngtcp2_strerror(error_code));
-            ngtcp2_conn_shutdown_stream(conn, stream_id, error_code);
-        }
-        if (is_shutdown)
-            data_callback = {};
+        conn.quic_manager->call([this, error_code]() {
+            log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
```
dr7ana, line 73: small nitpick, but can you move these trace logging function names to be between the async call and the function header? thinking in case it doesn't print the correct function name since its inside a callable

```
void Stream::close(uint64_t error_code)
{
        conn.quic...
```

tewinget: ~~`__PRETTY_FUNCTION__` is evaluated by the pre-processor at compile time which has no idea that the string is being put in a function object to be called later, so it will have the correct function name.  I put it inside the lambda so it prints when the meat of the function executes rather than when it's queued, as logs could get confusing then.~~

src/network.cpp
tewinget: I'm probably gonna just get rid of this call to `walk` entirely as it seems kinda jank and currently does nothing for us.

src/stream.cpp
jagerman: __PRETTY_FUNCTION__ is going to be ugly because it's in a lambda, but should still have the outer function name in it.  So still functional, but ugly.

src/network.cpp
tewinget: (and then just make sure we keep track of and clean up uvw handles appropriately rather than trying to get it to automagically keep track of them)

src/stream.cpp
jagerman: Maybe you want *both* traces?  i.e. so you trace when it's called, and then trace when it's actually called inside the event loop thread.

tewinget: Changed to use `source_location` and slightly custom `oxen::logging` invocation

Edit: also, it logs both when a function is queued and when it actually runs

src/network.cpp
tewinget: In fact, we kinda *need* to not rely on walk unless we can afford to use all our handles' `user_data` pointer solely to show ourselves that we own the handle, because we may be given an existing uv_loop and we can't trample the user's handles.

src/network.cpp
tewinget: const-ified, mildly begrudgingly

src/network.cpp
dr7ana: yeeeeup hopefully when we merge #8, we can just call something like this (edited to make more sense after our discussion):
```
ev_loop->walk([](uvw::udp_handle &&h){ 
    if (h.address in handler.our_{server,client}_addrs)
        h.close(); });
```
That's obviously pseudocode, but if we had a handy/inexpensive way to check the local address of the udp handle, maybe we could streamline udp_handle closing? Currently the addresses are keys in a map, and checking all keys is log(n), but maybe we can do something that's worst-case log(n)...?

src/client.cpp
```diff
@@ -59,11 +60,25 @@ namespace oxen::quic
...
         log::trace(log_cat, "Opening client stream...");
-        auto ctx = reinterpret_cast<ClientContext*>(context.get());
 
-        auto conn = get_conn(ctx->conn_id);
+        std::promise<std::shared_ptr<Stream>> p;
+        auto f = p.get_future();
+        handler->call([&data_cb, &close_cb, &p, this]() {
+            try
+            {
+                auto ctx = reinterpret_cast<ClientContext*>(context.get());
```
jagerman: Delete this and just use `context` instead of `ctx` in the next line.

src/stream.cpp
```diff
@@ -69,22 +69,24 @@ namespace oxen::quic
...
+                log::info(log_cat, "Stream is already shutting down");
+            else if (is_closing)
+                log::debug(log_cat, "Stream is already closing");
+            else
+            {
+                is_closing = is_shutdown = true;
+                log::info(log_cat, "Closing stream (ID: {}) with error code {}", stream_id, ngtcp2_strerror(error_code));
+                ngtcp2_conn_shutdown_stream(conn, 0, stream_id, error_code);
+            }
+            if (is_shutdown)
```
jagerman, line 85: perhaps: `if (is_shutdown && data_callback)`?

src/stream.cpp
```diff
@@ -122,8 +124,10 @@ namespace oxen::quic
 
     void Stream::when_available(unblocked_callback_t unblocked_cb)
     {
-        log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
-        unblocked_callbacks.push(std::move(unblocked_cb));
+        conn.quic_manager->call([this, unblocked_cb]() {
+            log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+            unblocked_callbacks.push(std::move(unblocked_cb));
```
jagerman: There is some implicit constness here that is mucking things up (not worth changing because this function is going away) but:
- lambda value captures are const by default (declare the lambda `mutable` to fix that)
- because it's const, the `std::move(unblocked_cb)` is a `const&&` rather than a `&&` rvalue reference, and that *can't* bind to the usual rvalue move operator, so will have to do a copy (and so the `std::move` does nothing, but is subtly misleading).

jagerman (approved):

src/stream.cpp
jagerman: Also: the `unblocked_cb` capture should be `unblocked_cb=std::move(unblocked_cb)` to more it into the lambda captures.

PR #5, opened 2023-06-02
Add chunked send interface

jagerman:
This allows you to queue up a callback to be invoked multiple times to be able to send chunks via callbacks that get automatically invoked as each chunk finishes being used by the stream buffer.

It also includes an (optional) final callback to invoke when the chunks are done so that you can know when it is safe to append post-chunk data to the stream.

See the added test code for an example of both.

### DIFF
--- a/include/quic/stream.hpp
+++ b/include/quic/stream.hpp
@@ -118,6 +118,96 @@ namespace oxen::quic
             send(std::basic_string_view<Char>{buf.data(), buf.size()}, std::make_shared<std::vector<Char>>(std::move(buf)));
         }
 
+      private:
+        // Implementations classes for send_chunks()
+
+        // chunk_sender: When sending chunks we construct *one* of these, then share its ownership
+        // across all the chunks in flight.  When each individual chunk gets destroyed, it called
+        // back into this to queue the next chunk, which this class then sends into the stream.
+        template <typename Container>
+        struct chunk_sender : std::enable_shared_from_this<chunk_sender<Container>>
+        {
+            using chunk_callback_t = std::function<Container()>;
+            using done_callback_t = std::function<void()>;
+
+            template <typename... Args>
+            static void make(int initial_queue, Args&&... args)
+            {
+                std::shared_ptr<chunk_sender<Container>> cs{new chunk_sender<Container>(std::forward<Args>(args)...)};
+                for (int i = 0; i < initial_queue; i++)
+                    cs->queue_next_chunk();
+            }
+
+          private:
+            // This is instantiated for each chunk, contains the chunk data itself, and is what we
+            // put into the keep_alive; during destruction, we queue the next chunk.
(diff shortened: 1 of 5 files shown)


include/quic/stream.hpp
```diff
@@ -95,22 +94,159 @@ namespace oxen::quic
...
+
+        // chunk_sender: When sending chunks we construct *one* of these, then share its ownership
+        // across all the chunks in flight.  When each individual chunk gets destroyed, it called
+        // back into this to queue the next chunk, which this class then sends into the stream.
+        //
+        // Container -- can be a value or a pointer, but not a reference.
+        template <typename Container>
+        struct chunk_sender : std::enable_shared_from_this<chunk_sender<Container>>
+        {
+            static_assert(!std::is_reference_v<Container>, "chunk_sender requires a value or pointer, not a reference");
```
dr7ana, line 132: should we throw instead?

include/quic/stream.hpp
```diff
@@ -95,22 +94,159 @@ namespace oxen::quic
...
+
+              public:
+                single_chunk(chunk_sender& cs, Container&& d) : _chunks{cs.shared_from_this()}, _data{std::move(d)} {}
+                ~single_chunk() { _chunks->queue_next_chunk(); }
+
+                bstring_view view() const
+                {
+                    if constexpr (is_pointer)
+                    {
+                        static_assert(sizeof(*_data->data()) == 1, "chunk_sender requires bytes data");
```
dr7ana, line 166: static_assert -> throw instead?

include/quic/stream.hpp
jagerman: When you can `static_assert` (i.e. can check conditions at compile time) that's always a better solution to fail to compile when preconditions aren't met.

include/quic/stream.hpp
jagerman: See above.

include/quic/stream.hpp
```diff
@@ -95,22 +94,159 @@ namespace oxen::quic
...
+                        static_assert(sizeof(*_data.data()) == 1, "chunk_sender requires bytes data");
+                        return {reinterpret_cast<const std::byte*>(_data.data()), _data.size()};
+                    }
+                }
+            };
+
+            chunk_sender(Stream& s, chunk_callback_t next, done_callback_t done) :
+                    str{s}, next_chunk{std::move(next)}, done{std::move(done)}
+            {
+                assert(next_chunk);
```
dr7ana, line 180: static_assert?

jagerman: Not compile-time checkable.

If this is false here, the caller did something wrong enough that it isn't something we want to be catchable.

dr7ana:
lgtm, just need to change the additions to `tests/CMakeLists.txt` to reflect merging of #6

PR #7, opened 2023-06-06
Batched Packet Sending

dr7ana:
- packets batched in groups of 8 to reduce syscall overhead
- packet send logic moved to Endpoint from Connection

### DIFF
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -34,8 +34,9 @@ namespace oxen::quic
 
         int init(ngtcp2_settings& settings, ngtcp2_transport_params& params, ngtcp2_callbacks& callbacks);
 
-        std::array<std::byte, NGTCP2_MAX_UDP_PAYLOAD_SIZE> send_buffer{};
+        send_buffer_t send_buffer{};
         size_t send_buffer_size = 0;
+        size_t n_packets = 0;
         ngtcp2_pkt_info pkt_info{};
 
       public:
@@ -89,7 +90,7 @@ namespace oxen::quic
 
         void on_io_ready();
 
-        io_result send();
+        io_result send(uint64_t ts);
 
         void flush_streams();
 
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -254,6 +254,8 @@ namespace oxen::quic
     io_result Connection::send(uint64_t ts)
     {
         log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+        if (n_packets == 0)
+            return io_result{0};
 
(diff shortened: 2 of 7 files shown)


include/quic/connection.hpp
```diff
@@ -34,8 +34,10 @@ namespace oxen::quic
 
         int init(ngtcp2_settings& settings, ngtcp2_transport_params& params, ngtcp2_callbacks& callbacks);
 
-        std::array<std::byte, NGTCP2_MAX_UDP_PAYLOAD_SIZE> send_buffer{};
+        static constexpr size_t batch_size = 8;
+        std::array<std::pair<std::array<std::byte, NGTCP2_MAX_PMTUD_UDP_PAYLOAD_SIZE>, size_t>, batch_size> send_buffer{};
```
jagerman: This type is unwieldy enough (and used in multiple places) that it ought to be a `using send_buffer_t = std::array<...` typedef.

src/connection.cpp
```diff
@@ -488,12 +487,16 @@ namespace oxen::quic
...
             }
 
-            log::info(log_cat, "Sending data packet with non-stream data frames");
-            if (auto rv = send_packet(nwrite); rv != 0)
-                return;
-            ngtcp2_conn_update_pkt_tx_time(conn.get(), ts);
+            buf.second = nwrite;
+
+            if (++n_packets == batch_size)
+                break;
```
jagerman: This doesn't seem right: this `for(;;)` loop may want to send more than 1 packet, but this loop could be restricted to just one iteration (if the previous loop filled up 7/8 slots).  I think we need the same append-and-send-if-full logic as in the upper loop.

src/endpoint.cpp
```diff
@@ -257,18 +257,53 @@ namespace oxen::quic
...
 
-        return io_result{0};
+        std::array<uv_buf_t, batch_size> raw_bufs;
+
+        log::info(log_cat, "Sending udp batch to {}:{}...", p.remote.ip.c_str(), p.remote.port);
+
+        char* packet_data_pos = packet_data->data();
+        for (int i = 0; i < n_pkts; ++i)
+        {
+            assert(bufs[i].second > 0);
```
tewinget:
```suggestion
            assert(buf[i].second > 0);
```

tewinget:
test 004 aborts and 005 fails, but both pass just fine if I rebase my pending `thread_safety` branch on top of this so apart from the suggested change in `endpoint.cpp` I'd say this is good to go.

jagerman:
Closing (because this is embedded into #1, which takes it much further for performance optimizations)

PR #9, opened 2023-06-21
Max stream opts

dr7ana:
- user can pass max stream count from top-level
- created config_t struct to store future user config values
- augmented [004] to test configuration

### DIFF
--- a/include/quic/opt.hpp
+++ b/include/quic/opt.hpp
@@ -13,18 +13,29 @@ namespace oxen::quic
         {
             using Address::Address;
         };
+
         struct remote_addr : public Address
         {
             using Address::Address;
         };
+
+        struct max_streams
+        {
+            int stream_count = DEFAULT_MAX_BIDI_STREAMS;
+            max_streams() = default;
+            explicit max_streams(int s) : stream_count(s) {};
+        };
+
         struct remote_tls : public GNUTLSCert
         {
             using GNUTLSCert::GNUTLSCert;
         };
+
         struct local_tls : public GNUTLSCert
         {
             using GNUTLSCert::GNUTLSCert;
         };
+
         struct server_tls : public GNUTLSCert
         {
             explicit server_tls(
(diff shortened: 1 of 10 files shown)


include/quic/opt.hpp
```diff
@@ -13,18 +13,28 @@ namespace oxen::quic
...
         };
+
         struct remote_addr : public Address
         {
             using Address::Address;
         };
+
+        struct max_streams
+        {
+            int stream_count{};
```
jagerman: Maybe default this value to the constexpr constant?  That way passing a default-constructed `max_streams` would be the same as not passing it at all.

tests/004-streams.cpp
```diff
@@ -70,4 +70,73 @@ namespace oxen::quic::test
...
+        opt::client_tls client_tls{"./clientkey.pem"s, "./clientcert.pem"s, "./servercert.pem"s};
+
+        opt::local_addr server_local{"127.0.0.1"s, static_cast<uint16_t>(5500)};
+        opt::local_addr client_local{"127.0.0.1"s, static_cast<uint16_t>(4400)};
+        opt::remote_addr client_remote{"127.0.0.1"s, static_cast<uint16_t>(5500)};
+
+        log::debug(log_cat, "Calling 'server_listen'...");
+        auto server = test_net.server_listen(server_local, server_tls, max_streams);
+
+        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
jagerman: This would be a lot less verbose as `1s` instead of `std::chrono::milliseconds(1000)`

tests/004-streams.cpp
```diff
@@ -70,4 +70,73 @@ namespace oxen::quic::test
         REQUIRE(conn->pending_streams.size() == 1);
```
jagerman: It's probably worth updating this test to use fewer streams now (we were using 32 only because it was the hard-coded value before).  Maybe construct it with 10 max streams (and then reduce the size-`36` vector to size-`14`).

jagerman: (The comment refers to the whole preceeding section, not this specific line).

jagerman (approved):

tests/004-streams.cpp
jagerman: Scratch that, that's what the test below does.  We can probably just use the new version and drop the old version.

PR #12, opened 2023-06-26
Endpoint transpose

dr7ana:
PR for review and discussion on restructuring the endpoints


tests/001-handshake.cpp
```diff
@@ -13,42 +13,76 @@ namespace oxen::quic::test
...
+
+        auto client_tls = GNUTLSCreds::make("./clientkey.pem"s, "./clientcert.pem"s, "./servercert.pem"s);
+		client_tls->client_tls_policy = std::move(outbound_tls_cb);
+
+        opt::local_addr server_local{"127.0.0.1"s, 5500};
+        opt::local_addr client_local{"127.0.0.1"s, 4400};
+        opt::remote_addr client_remote{"127.0.0.1"s, 5500};
+
+		auto server_endpoint = test_net.endpoint(server_local);
+		bool sinit = server_endpoint->inbound_init(server_tls, server_data_cb, stream_open_cb);
```
jagerman: This name is a bit awkward; how about just `listen`?

include/quic/endpoint.hpp
```diff
@@ -109,27 +188,10 @@ namespace oxen::quic
         // Accepts new connection, returning either a ptr to the Connection
         // object or nullptr if error. Virtual function returns nothing --
```
tewinget: comment is out-of-sync with changes

include/quic/endpoint.hpp
```diff
@@ -109,27 +188,10 @@ namespace oxen::quic
...
-        // - ecn value [1 byte]: provided by ngtcp2 (Only the lower 2 bits are actually used).
-        //
-        // \param psuedo_port - the remote's pseudo-port (will be 0 if the remote is a
-        //      server, > 0 for a client remote)
-        // \param ecn - the ecn value from ngtcp2
-        //
-        // Returns the number of bytes written to buf
-        virtual size_t write_packet_header(uint16_t pseudo_port, uint8_t ecn) { return 0; }
+	  private:
+	  	//
```
tewinget: a comment nobody can read it perhaps a bit *too* private...

include/quic/utils.hpp
```diff
@@ -207,6 +201,11 @@ namespace oxen::quic
         return numeric_host_family(hostname, AF_INET) || numeric_host_family(hostname, AF_INET6);
     }
 
+	enum Direction
```
tewinget: This works, but...it's basically just a `bool` with extra steps, no?

include/quic/network.hpp
```diff
@@ -12,172 +12,96 @@ extern "C"
...
         Network();
         ~Network();
 
+		std::shared_ptr<Endpoint> endpoint(Address& local_addr);
+        std::shared_ptr<Endpoint> endpoint(const Address& local_addr);
+
+        void close();
+
+		// Find and return the endpoint with the given local address; returns nullptr if not found
+		Endpoint* get_endpoint(const Address& local);
```
tewinget: should probably return `shared_ptr` as the functions above do.

src/network.cpp
```diff
@@ -43,6 +56,20 @@ namespace oxen::quic
         close();
     }
 
+	std::shared_ptr<Endpoint> Network::endpoint(const Address &local_addr)
+	{
+		return endpoint(const_cast<Address&>(local_addr));
```
jagerman: This cast is dangerous, but also seems unnecessary; why do we ever need to take the address as a non-const reference?

tests/001-handshake.cpp
```diff
@@ -13,42 +13,76 @@ namespace oxen::quic::test
...
+
+		REQUIRE(sinit);
+
+        auto client_endpoint = test_net.endpoint(client_local);
```
jagerman, line 58: I had to look up whether this call would allow me to re-use the same address for multiple endpoints (it doesn't).  So either:

1. We should make it recognize that you are reusing the address of an existing endpoint and just return that; or

2. We should rename it so that it's clear that this strictly creates a new one (e.g. `make_endpoint` or `start_endpoint` or `bind` or something along those lines).

include/quic/endpoint.hpp
```diff
@@ -20,52 +20,130 @@ extern "C"
...
+
+			return f.get();
+		};
 
         std::shared_ptr<uvw::loop> get_loop();
 
-        void handle_packet(Packet& pkt);
-
-        void close_connection(Connection& conn, int code = NGTCP2_NO_ERROR, std::string_view msg = "NO_ERROR"sv);
+        Connection* get_conn_ptr(ConnectionID ID);		// query by conn ID
```
tewinget: This and the one below should be private so library users don't think they should be calling them.

include/quic/connection.hpp
```diff
@@ -165,6 +196,30 @@ namespace oxen::quic
         }
     };
 
+	class connection_interface
+    {
```
jagerman: I don't think this should be a concrete class, but rather a pure virtual interface, along the lines of:

```C++
class connection_interface {
    public:
    	std::shared_ptr<Stream> get_new_stream(
                stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr) = 0;
};
```

which is then implemented by `Connection`, but we always return the connection wrapped in a `shared_ptr<connection_interface>`.

include/quic/utils.hpp
jagerman:
```suggestion
	enum class Direction
```

include/quic/endpoint.hpp
dr7ana: its our little secret tom

include/quic/utils.hpp
```diff
@@ -284,32 +283,39 @@ namespace oxen::quic
...
             return reinterpret_cast<const sockaddr_in6&>(_sock_addr);
         }
 
-        uint16_t port() const
+		inline uint16_t port_horder() const
+        {
+            assert(is_ipv4() || is_ipv6());
+            return is_ipv4() ? ntohs(reinterpret_cast<const sockaddr_in&>(_sock_addr).sin_port)
+                             : reinterpret_cast<const sockaddr_in6&>(_sock_addr).sin6_port;
+        }
```
jagerman, line 317: Don't need this; instead just fix `port()` to do this, but using `oxenc::big_to_host` to do the endian conversion.  Will be something like:
```C++
#include <oxenc/endian.h>
// ...
            return oxenc::big_to_host(
                is_ipv4() ? reinterpret_cast<const sockaddr_in&>(_sock_addr).sin_port
                             : reinterpret_cast<const sockaddr_in6&>(_sock_addr).sin6_port);
```

include/quic/network.hpp
```diff
@@ -12,172 +12,96 @@ extern "C"
...
-                {
-                    p.set_exception(std::current_exception());
-                }
-            });
-
-            return f.get();
-        }
+        // Maps local listening address to respective endpoint
+        std::map<Address, std::shared_ptr<Endpoint>, addr_comparator> endpoint_map;
+        std::map<Address, std::shared_ptr<uv_udp_t>, addr_comparator> handle_map;
```
jagerman: This feels out of place for a `map` here as `Address` isn't something that is inherently orderable (and the order applied here is applying dictionary ordering on the stringified form of the address).

Is there a reason to avoid using an `unordered_map` here?

include/quic/utils.hpp
```diff
@@ -378,6 +384,16 @@ namespace oxen::quic
...
 
+	struct addr_comparator
+	{
+		using is_transparent = std::true_type;
+
+		inline bool operator()(const Address& lhs, const Address& rhs) const
+		{
+			return lhs.to_string() < rhs.to_string();
+		}
+	};
```
jagerman: `to_string` is going to be quite expensive because of the need to allocate two separate strings for every comparison (which means, to do any map lookup or insertions, you will be doing ~ 2*log(k) string allocations for each lookup/insert).

include/quic/network.hpp
dr7ana: It's unnecessary now, so deleting it entirely. I changed `udp_data` to store an `Endpoint&` rather a shared_ptr to the handler which eliminated the call to `get_endpoint`

include/quic/connection.hpp
dr7ana: completely agree, done and implemented

tests/001-handshake.cpp
dr7ana: Sounds good. Implemented option 1 to update this PR until we discuss details on the alternate function calls for this and the others

include/quic/endpoint.hpp
```diff
@@ -22,51 +22,117 @@ extern "C"
...
 #include "connection.hpp"
 #include "context.hpp"
+#include "network.hpp"
 #include "utils.hpp"
 
 namespace oxen::quic
 {
-    class Connection;
-    class Handler;
+    using conn_ptr_pair = std::pair<std::shared_ptr<Connection>, std::shared_ptr<connection_interface>>;
```
jagerman: Seems dead now?

include/quic/connection.hpp
```diff
@@ -20,62 +20,113 @@ namespace oxen::quic
...
+                stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr) = 0;
+    };
+
+    class Connection : public connection_interface, public std::enable_shared_from_this<Connection>
+    {
+      protected:
+        friend class Network;
+        friend class Endpoint;
+
+        using conn_ptr_pair = std::pair<std::shared_ptr<Connection>, std::shared_ptr<connection_interface>>;
```
jagerman: Duplicate typedef, and seems dead now.

include/quic/connection.hpp
```diff
@@ -20,62 +20,111 @@ namespace oxen::quic
...
+                const ConnectionID& scid,
+                const ConnectionID& dcid,
+                const Path& path,
+                std::shared_ptr<uv_udp_t> handle,
+                std::shared_ptr<TLSCreds> creds,
+                config_t u_config)
+        {
+            log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+            return _make_conn(ep, scid, dcid, path, handle, std::move(creds), u_config, Direction::OUTBOUND);
+        };
```
jagerman: These functions seems verbose; would it be easier to just expose `make_conn` as a public static method, since all these do is simply forward all their arguments (plus a Direction) to it?  (And because this is a quasi-internal function, since Connection isn't really user-facing).

dr7ana: So currently `make_{inbound,outbound}_conn` is protected, but the `_make_conn` is private. There's definitely an argument for bypassing the inbound/outbound functions entirely, but what purpose would making `_make_conn` public serve? Currently it's only called by `Endpoint`, which has friend access

dr7ana: Deleting `make_{inbound,outbound}_conn` and making `make_conn` private per this morning's discussion

Motivation: gating access and removing `friend` class specifiers in adding things to `connection_interface`

include/quic/endpoint.hpp
```diff
@@ -17,56 +17,118 @@ extern "C"
...
+        std::shared_ptr<uv_udp_t> handle;
+        bool accepting_inbound{false};
+        Network& net;
 
-        std::shared_ptr<Handler> handler;
+      public:
+        explicit Endpoint(Network& n, const Address& listen_addr);
+
+        template <typename... Opt>
+        bool listen(Opt&&... opts)
```
jagerman, line 46: This function can only ever return true (it throws on failure), so shouldn't be a `bool` return.

include/quic/endpoint.hpp
```diff
@@ -87,8 +149,8 @@ namespace oxen::quic
         //      a short period of time allowing any lagging packets to be caught
         //
         //      They are indexed by connection ID, storing the removal time as a uint64_t value
-        //
-        std::unordered_map<ConnectionID, std::unique_ptr<Connection>> conns;
+        std::unordered_map<ConnectionID, std::shared_ptr<Connection>> conns;
+
         std::queue<std::pair<ConnectionID, uint64_t>> draining;
```
jagerman: I think this queue should be a `map<expiry, ConnectionID>` instead because I believe the draining period differs per connection (so a low-latency connection could get its drain period unintentionally extended because it gets stuck behind a high-latency drain).

src/endpoint.cpp
```diff
@@ -213,17 +218,68 @@ namespace oxen::quic
...
+                    p.set_value(it->second.get());
+                    return;
+                });
+
+                return f.get();
+            }
+        };
+    };
+
     void Endpoint::handle_conn_packet(Connection& conn, Packet& pkt)
```
jagerman, line 271: This seems misplaced: everything in here, I think, could be in a Connection method.

include/quic/endpoint.hpp
```diff
@@ -17,56 +17,118 @@ extern "C"
...
+                }
+            });
+
+            return f.get();
+        };
 
         std::shared_ptr<uvw::loop> get_loop();
 
+        // query a list of all active inbound and outbound connections paired with a conn_interface
+        std::list<std::pair<ConnectionID, std::shared_ptr<connection_interface>>> get_all_conns(std::optional<Direction> d = std::nullopt);
```
jagerman: I think this can just return a `list<shared_ptr<connection_interface>>`: you should be able to retrieve the ConnectionID out of the connection_interface.

include/quic/connection.hpp
```diff
@@ -13,137 +13,157 @@
...
+        { return _source_cid; }
+
+        Direction direction()
+        { return dir; }
+
+        bool is_closing()
+        { return closing; }
+        bool is_draining()
+        { return draining; }
+        const ConnectionID scid()
```
jagerman: Since ConnectionID is non-trivial, better to return these as `const ConnectionID&` references so that we don't have to copy when doing simple access.  (Likewise for dcid(), path()).

jagerman: and `get_conn_id()`

include/quic/connection.hpp
```diff
@@ -13,137 +13,157 @@
...
+        void io_ready();
+
+        const GNUTLSSession* get_session()
+        { return dynamic_cast<GNUTLSSession*>(tls_session.get()); };
+
+        std::shared_ptr<Stream> get_new_stream(
+                stream_data_callback_t data_cb = nullptr, stream_close_callback_t close_cb = nullptr) override;
+
+        ConnectionID get_conn_id() override
+        { return _source_cid; }
```
jagerman: What's the difference between this and `scid()`?

include/quic/endpoint.hpp
dr7ana: Do we want to have a bool return to check if the server was initialized correctly? If so, I can have the promise set false and return the exception to console.

include/quic/connection.hpp
```diff
@@ -13,137 +13,150 @@
...
+                Direction dir,
+                ngtcp2_pkt_hd* hdr = nullptr);
+
+        // Returns a pointer to the owning endpoint, else nullptr
+        Endpoint* endpoint();
+        const Endpoint* endpoint() const;
+
+        void io_ready();
+
+        const GNUTLSSession* get_session()
```
tewinget: Connection has no idea which sort of TLSSession it owns and should not care; that's the purpose of abstracting TLSCreds and TLSSession.

Ok right *now* it knows it's a GNUTLSSession but that's because that's the only one implemented...in general it shouldn't know or care.

include/quic/connection.hpp
```diff
@@ -165,6 +190,10 @@ namespace oxen::quic
         {
             return conn.get();
         }
+
+        // returns number of currently pending streams for use in test cases
+        size_t num_pending()
```
tewinget: could be const I guess?

PR #13, opened 2023-06-28
Replace libuv with libevent

jagerman:
Builds on top of #12.

### DIFF
--- a/include/quic/endpoint.hpp
+++ b/include/quic/endpoint.hpp
@@ -10,18 +10,22 @@ extern "C"
 #include <ngtcp2/ngtcp2_crypto_gnutls.h>
 }
 
+#include <event2/event.h>
+
 #include <cstddef>
+#include <list>
 #include <memory>
 #include <numeric>
 #include <optional>
 #include <queue>
 #include <random>
 #include <string>
-#include <uvw.hpp>
+#include <unordered_map>
 
 #include "connection.hpp"
 #include "context.hpp"
 #include "network.hpp"
+#include "udp.hpp"
 #include "utils.hpp"
 
 namespace oxen::quic
@@ -34,8 +38,8 @@ namespace oxen::quic
 
       private:
         const Address local;
-        std::shared_ptr<uvw::timer_handle> expiry_timer;
-        std::shared_ptr<uv_udp_t> handle;
+        event_ptr expiry_timer;
+        std::unique_ptr<UDPSocket> socket;
         bool accepting_inbound{false};
         Network& net;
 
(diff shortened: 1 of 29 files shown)


include/quic/endpoint.hpp
```diff
@@ -108,17 +111,21 @@ namespace oxen::quic
             return f.get();
         }
 
-        std::shared_ptr<uvw::loop> get_loop();
+        const std::shared_ptr<event_base>& get_loop() { return net.loop(); }
```
tewinget, line 114: `const std::shared_ptr<event_base>& get_loop() const { return net.loop(); }`

include/quic/endpoint.hpp
```diff
@@ -108,17 +111,21 @@ namespace oxen::quic
...
+
+        const std::unique_ptr<UDPSocket>& get_socket() { return socket; }
```
tewinget, line 116: ^ see above

src/connection.cpp
```diff
@@ -566,29 +539,15 @@ namespace oxen::quic
...
         stream->set_ready();
 
-        stream->stream_id = id;
-        uint64_t rv{0};
+        log::debug(log_cat, "Local endpoint creating stream to match remote");
 
-        auto ep = stream->conn.endpoint();
-
-        if (ep)
+        if (uint64_t app_err_code = context->stream_open_cb ? context->stream_open_cb(*stream) : 0; app_err_code != 0)
```
tewinget, line 547: I believe the `return NGTCP2_ERR_CALLBACK_FAILURE` below is connection-fatal.  Commenting to remind us to consider whether we want a stream cb failure to be connection-fatal.

tewinget: Since `app_err_code` is ours to define, we may want some which are and some which are not connection-fatal.

src/connection.cpp
```diff
@@ -749,7 +711,7 @@ namespace oxen::quic
 
         ngtcp2_settings_default(&settings);
 
-        settings.initial_ts = get_timestamp();
+        settings.initial_ts = get_timestamp().count();
```
tewinget, line 714: in `on_io_ready` you changed `get_timestamp` to `get_time`, would that change be appropriate here, or this change there instead?

include/quic/endpoint.hpp
jagerman: This is returning a non-const event_base pointer (just wrapped in a const lvalue reference to the shared_ptr that holds it); the event_base it is returning, however, is non-const.

include/quic/endpoint.hpp
jagerman: ^ see above

src/connection.cpp
jagerman: `get_time` returns a steady_clock::time_point, this returns the .time_since_epoch() of that time_point.  (The main difference is that it now returns typesafe nanoseconds instead of a bare integer).

src/connection.cpp
jagerman: Fixed.

tewinget (approved):

PR #14, opened 2023-06-29
Warnings

jagerman:
- Enable additional compilation warnings to catch various common errors
- Add `-DWARNINGS_AS_ERRORS=ON` to make all warnings fatal errors
- Fix resulting compiler warnings in the library and test suite
- Fix a potential bug allowing copying/moving of Endpoint/Connection/Stream objects.  (We always hold these in shared_ptrs that we copy or move, but don't expect the instances themselves to copy or move).

### DIFF
--- a/include/quic/crypto.hpp
+++ b/include/quic/crypto.hpp
@@ -17,6 +17,7 @@ namespace oxen::quic
     {
       public:
         virtual std::unique_ptr<TLSSession> make_session(const ngtcp2_crypto_conn_ref& conn_ref, bool is_client) = 0;
+        virtual ~TLSCreds() = default;
     };
 
     class TLSSession
--- a/include/quic/endpoint.hpp
+++ b/include/quic/endpoint.hpp
@@ -204,7 +204,7 @@ namespace oxen::quic
 
         void check_timeouts();
 
-        Connection* accept_initial_connection(const Packet& pkt, const ConnectionID& dcid);
+        Connection* accept_initial_connection(const Packet& pkt);
     };
 
 }  // namespace oxen::quic
--- a/include/quic/endpoint.hpp
+++ b/include/quic/endpoint.hpp
@@ -36,7 +36,6 @@ namespace oxen::quic
         friend class Connection;
         friend class Stream;
 
-      private:
         const Address local;
         event_ptr expiry_timer;
         std::unique_ptr<UDPSocket> socket;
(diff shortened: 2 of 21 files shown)


src/connection.cpp
```diff
@@ -140,19 +140,19 @@ namespace oxen::quic
...
 
-    int recv_rx_key(ngtcp2_conn* conn, ngtcp2_encryption_level level, void* user_data)
+    int recv_rx_key(ngtcp2_conn* /*conn*/, ngtcp2_encryption_level /*level*/, void* /*user_data*/)
     {
         // fix this
         return 0;
     }
 
-    int recv_tx_key(ngtcp2_conn* conn, ngtcp2_encryption_level level, void* user_data)
+    int recv_tx_key(ngtcp2_conn* /*conn*/, ngtcp2_encryption_level /*level*/, void* /*user_data*/)
```
tewinget, line 149: function itself currently unused, behavior currently facilitated by the ngtcp2-via-gnutls implementation

src/connection.cpp
```diff
@@ -140,19 +140,19 @@ namespace oxen::quic
         return 0;
     }
 
-    int recv_rx_key(ngtcp2_conn* conn, ngtcp2_encryption_level level, void* user_data)
+    int recv_rx_key(ngtcp2_conn* /*conn*/, ngtcp2_encryption_level /*level*/, void* /*user_data*/)
```
tewinget, line 143: function itself currently unused, behavior currently facilitated by the ngtcp2-via-gnutls implementation

include/quic/endpoint.hpp
```diff
@@ -36,14 +36,19 @@ namespace oxen::quic
         friend class Connection;
         friend class Stream;
 
-      private:
```
tewinget, line 39: private *is* default for classes, but I don't see an issue with being explicit about it.  That said, I don't have strong feelings either way.

include/quic/crypto.hpp
```diff
@@ -17,6 +17,7 @@ namespace oxen::quic
     {
       public:
         virtual std::unique_ptr<TLSSession> make_session(const ngtcp2_crypto_conn_ref& conn_ref, bool is_client) = 0;
+        virtual ~TLSCreds() = default;
```
tewinget, line 20: I was about to say the same applies to `TLSSession` but then I expanded context here and saw that I did put one with a trace log to know when it happens...

src/connection.cpp
dr7ana: @tewinget will this be needed for the null crypto implementation you're working on? (same question for your comment on `recv_rx_key` below)

tewinget: Something to this effect will be, yes.

dr7ana (approved):

PR #16, opened 2023-07-01
Drone CI, static builds, Windows builds

jagerman:
- Adds CI for the project.
- Adds static dep builds
- Makes windows build work in mingw (though doesn't appear to *work* yet).
- Fixes various problems caught by the CI jobs.

This adds a suite of Drone CI builds, and fixes various issues where things weren't building.

This includes full static builds, via the same -DBUILD_STATIC_DEPS=ON mode we have in other software.

Finally this adds a Windows build, including lots of ugly unholiness brought in from hell, including a suitably named in constexpr constant to help tell when we are in such an environment.


include/quic/gnutls_crypto.hpp
```diff
@@ -33,46 +33,46 @@ namespace oxen::quic
...
-        // returns truew if path is not empty OR mem has a value set
-        inline explicit operator bool() const noexcept { return (!path.empty() || mem.size); }
+        // returns true if path is not empty OR mem has a value set
+        explicit operator bool() const noexcept { return (!path.empty() || mem.size); }
 
-        operator const char*() { return path.c_str(); }
-        operator const gnutls_datum_t*() { return &mem; }
-        operator gnutls_x509_crt_fmt_t() { return ext; }
+        // Hidden behind a template so that implicit conversion to pointer doesn't cause trouble
+        template <typename T, typename = std::enable_if_t<std::is_same_v<T, gnutls_datum_t>>>
```
dr7ana, line 71: This was a good idea. Was there a specific issue prompting this change, or is this prophylactic?

include/quic/utils.hpp
```diff
@@ -57,6 +61,13 @@ namespace oxen::quic
     using bstring_view = std::basic_string_view<std::byte>;
     namespace log = oxen::log;
 
+    constexpr bool IN_HELL =
```
dr7ana, line 64: 10/10

include/quic/utils.hpp
```diff
@@ -455,27 +466,52 @@ namespace oxen::quic
         explicit io_result(int errno_val) : error_code{errno_val} {}
 
         // Constructs an io_result with an ngtcp2 error value.
-        io_result(int err, ngtcp2_error_code_t) : error_code{err}, is_ngtcp2{true} {}
+        static io_result ngtcp2(int err)
```
dr7ana, line 469: Minor: maybe renaming these to something like `ngtcp2_err` and `wsa_err` would be a bit clearer?

include/quic/utils.hpp
```diff
@@ -583,7 +619,7 @@ namespace std
             }
 
             auto h = hash<string_view>{}(addr_data);
-            h ^= hash<in_port_t>{}(port) + inverse_golden_ratio + (h << 6) + (h >> 2);
+            h ^= hash<decltype(port)>{}(port) + inverse_golden_ratio + (h << 6) + (h >> 2);
```
dr7ana, line 622: Educational question: Previously the hash type was explicit to match the declaration of `port`. Why the change to `decltype`? Is this for platform compatibility?

dr7ana (commented):
looks great, left a few questions about things I thought i was curious about

include/quic/gnutls_crypto.hpp
jagerman: Not currently causing a problem AFAIK, but implicit conversion to a pointer type is all sorts of shotgun-y (because it allows things like `if (foo)` unintentionally).

include/quic/utils.hpp
jagerman: They are static functions, so they always have to be prefixed with the `io_result::ngtcp(123)`

include/quic/utils.hpp
jagerman: Because f-ing Windows has half-baked posix support for things like `in_port_t` that are supposed to be provided by aren't.

PR #17, opened 2023-07-02
Datagrams

### LINKED ISSUE #23: dual IPv6/IPv4 mode not working
In the test suite, binding a server to the default address (added in #17), then connecting a client to `127.0.0.1` on the server's port doesn't work: the client connects and gets packets to the server fine, but the *return* packets are being sent to `::ffff:127.0.0.1:PPP` (where PPP is the client's ephemeral port) and never arrive back at the client.

dr7ana:
Currently:
- datagram sending functional with no packet-splitting policy (see 007 test case)
- added template parameterization to Network::endpoint s.t. datagram support and packet splitting is enabled for all connections in/out of one endpoint
- ngtcp2 transport params populated by top level endpoint configurations
- 007 unit tests for setting and getting datagram parameters

TODO (maybe for a second PR?):
- send datagrams with packet-splitting
- tetris buffer


src/connection.cpp
```diff
@@ -766,6 +782,19 @@ namespace oxen::quic
         // config values
         params.initial_max_streams_bidi = (uconfig.max_streams) ? uconfig.max_streams : DEFAULT_MAX_BIDI_STREAMS;
 
+        if (uconfig.datagram_support)
+        {
+            params.max_datagram_frame_size = 65535;
```
jagerman, line 907: Deserves a comment along the lines of:
```
// This is effectively an "unlimited" value, which lets us accept any size that fits into a QUIC packet.
```
because otherwise I was confused by us not being able to accept anything remotely as large as this.

include/quic/connection.hpp
```diff
@@ -93,63 +143,146 @@ namespace oxen::quic
...
+        std::shared_ptr<IOContext> context;
+        Direction dir;
+        const ConnectionID _source_cid;
+        ConnectionID _dest_cid;
+        Path _path;
+        std::function<void(Connection&)> on_closing;  // clear immediately after use
+        const int _max_streams{DEFAULT_MAX_BIDI_STREAMS};
+        const bool _datagrams_enabled{false};
+        const bool _packet_splitting{false};
+        const Splitting _policy{Splitting::NONE};
```
tewinget, line 169: The variable name `_policy` here seems ambiguous.

src/connection.cpp
```diff
@@ -766,6 +782,19 @@ namespace oxen::quic
...
+            // default ngtcp2 values set by ngtcp2_settings_default_versioned
```
jagerman, line 908: If these are already set to these values by our calls to `ngtcp2_transport_params_default`/`ngtcp2_settings_default` then why set them again here (e.g. if the ngtcp2 defaults change, are we intentionally not willing to accept that change for some reason?)

include/quic/connection.hpp
dr7ana: I dunno. the enum class name is pretty clear what it is, so renaming it to `_splitting_policy` seems redundant. Also it's private and the getter is `Splitting Connection::packet_splitting_policy()`

include/quic/connection.hpp
```diff
@@ -93,63 +143,146 @@ namespace oxen::quic
...
+        datagram_recv_callback dgram_data_cb;
+
+        /// Datagram Numbering:
```
tewinget, line 172: I like how this is laid out.  I'd say that `packet` and `datagram` could be used more precisely here though, to avoid confusion.  For example, I'd say it should be phrased that we might split "packets" into multiple (two) datagrams.  Trying to cover all cases:

-----

> Each datagram ID is incremented by four from the previous one, regardless of whether we are splitting ~~datagrams~~ **packets**.

> In the unsplit packet scheme, the dgram ID of each ~~packet~~ **datagram** satisfies the rule:

> In the split-packet scheme, the dgram ID of the first of two ~~packets~~ **datagrams** satisfies the rule:

> The second of the two ~~packets~~ **datagrams** satisfies the rule:

> As a result, a packet-splitting endpoint should never send or receive a ~~packet~~ **datagram** whose ID is a perfect multiple of 4

src/connection.cpp
dr7ana: The only one i saw that `ngtcp2_settings_default` assigns is 

`settings->max_tx_udp_payload_size = 1500 - 48;`

I set it again explicitly in case we find the given overhead they estimated was insufficient and need to change the value, but I can absolutely remove it if its not helpful

The other one that `ngtcp2_transport_params_default` sets 

`params->max_udp_payload_size = NGTCP2_DEFAULT_MAX_RECV_UDP_PAYLOAD_SIZE;`

I set it to 65535 as per the rfc; their macro is 65527

include/quic/connection.hpp
dr7ana: Thank you for the edits! That all makes perfect sense, I'll make the change

include/quic/connection.hpp
tewinget: If the getter is used anywhere/everywhere it is referenced even within this class, that will be fine.  If not, we will later have something like:

```
// user datagram comes in
if (_policy)
{
    // split the datagram if needed
}
else
{
    // don't split
}
```

In that case, reading the code some time in the future it could be confusing/ambiguous what `_policy` means.  It would likely become clear quickly enough by reading the body of the `if` statement, but that feels suboptimal.

src/address.cpp
```diff
@@ -46,7 +46,7 @@ namespace oxen::quic
 
     std::string Path::to_string() const
     {
-        return "{{{} ➙ {}}}"_format(local, remote);
+        return "{{{} ➙ {}}}"_format(local.to_string(), remote.to_string());
```
jagerman: These should not be necessary; local and remote should already be formattable.

src/connection.cpp
```diff
@@ -303,23 +339,27 @@ namespace oxen::quic
...
-    bool Connection::send(pkt_tx_timer_updater* pkt_updater)
+    bool Connection::send(Send send_target, pkt_tx_timer_updater* pkt_updater)
     {
         log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
         assert(n_packets > 0 && n_packets <= MAX_BATCH);
 
         sent_counter += n_packets;
 
-        auto rv = endpoint().send_packets(remote(), send_buffer.data(), send_buffer_size.data(), send_ecn, n_packets);
+        auto rv = endpoint().send_packets(_path.remote, send_buffer.data(), send_buffer_size.data(), send_ecn, n_packets);
```
jagerman, line 317: Why make this change?

include/quic/connection.hpp
dr7ana: That's a fair point, ill go through and make sure that it uses the getter even internally

include/quic/context.hpp
```diff
@@ -12,73 +12,52 @@
...
 
+        ~IOContext() = default;
+
       private:
-        void handle_inbound_opt(std::shared_ptr<TLSCreds> tls);
-        void handle_inbound_opt(opt::max_streams ms);
-        void handle_inbound_opt(stream_data_callback_t func);
-        void handle_inbound_opt(stream_open_callback_t func);
-        void handle_inbound_opt(stream_close_callback_t func);
+        void handle_ioctx_opt(std::shared_ptr<TLSCreds> tls);
```
tewinget, line 57: I think this one should be explicit from `Endpoint` (as you did with `Direction`) and the Endpoint should require a non-null `TLSCreds` on construction.  i.e. an Endpoint should have one `TLSCreds` shared among all connections, inbound or outbound.

dr7ana: makes sense to me :+1:

jagerman (changes requested):
Overall comments:

Datagrams should use `NGTCP2_WRITE_DATAGRAM_FLAG_MORE` to allow packing small datagrams together into the same quic packet.  But this dovetails into the next point:

We don't want a separate loop & timer here for datagrams and stream data: we want, when feasible, to pack them together (e.g. if we have 100 bytes of stream data, and three 50-byte datagrams then we can fit all 4 things into one packet).

This ought to be not too difficult to implement with the current loop in `flush_streams` (which maybe we should rename to `flush_packets`):

- instead of appending one `nullptr` into the `strs` list we append *two* nullptrs.
- drop the `assert(stream || strs.empty())`
- inside the loop, if we get a nullptr and `strs` is *not* empty then it's the datagrams case (otherwise `strs` is empty so it's the final `-1` psuedo-stream case).
- for the datagrams case we need to skip if we have no datagrams to send, equivalent to what we do when a stream has no data to send in the `else if (bufs.empty()) { ... }` condition.
- if we're in the datagrams case then we need to call `ngtcp2_conn_writev_datagram` instead of `ngtcp2_conn_writev_stream`.  Most of the arguments will be the same, aside from the `ndatalen` and `stream_id` (which will be a `paccepted` and 0 for datagrams).
- the datagram return value is a bit wonky: unlike writev_stream, it can return a >0 final packet size but with `*paccepted = 0`, which indicates that the next datagram cannot be packed into the current packet: we need to fire off the just-finished packet, but then need to invoke the datagram handler again.
    - The easiest way to do this would be to re-add the `nullptr` stream pointer to the *beginning* of `strs` via a `push_front(nullptr)` instead of just-before-the-last, to avoid needing a duplicate writev_datagrams call, and then `continue;` the loop.
- At the end of the loop the current `if (!stream)` needs to be changed to `if (!stream && strs.empty())` but also we need the `else if` case right after than to also handle a "we have more datagrams" condition for re-appending `streams` to `strs`.

include/quic/gnutls_crypto.hpp
```diff
@@ -104,10 +129,10 @@ namespace oxen::quic
 
         void set_tls_hook_functions();  // TODO: which and when?
       public:
-        GNUTLSSession(GNUTLSCreds& creds, const ngtcp2_crypto_conn_ref& conn_ref_, bool is_client);
+        GNUTLSSession(GNUTLSCreds& creds, bool is_client);
         ~GNUTLSSession();
 
-        void* get_session() override { return session; };
+        void* get_session() override { return reinterpret_cast<gnutls_session_t>(session); };
```
tewinget: `session` is already a `gnutls_session_t`

dr7ana: oops, was checking something debugging a CI build. thanks

include/quic/stream.hpp
```diff
@@ -107,6 +85,44 @@ namespace oxen::quic
...
+        {
+            size_t sum{0};
+            if (user_buffers.empty())
+                return sum;
+            for (const auto& [data, store] : user_buffers)
+                sum += data.size();
+            return sum;
+        }
+
+        inline size_t unacked() const { return unacked_size; }
```
tewinget, line 118: We could expose this and `unsent` to the user I think.  They're const so it can't hurt, and I can see utility in the user having that information.

jagerman:
Also for sending datagrams, we want some feedback for the user in case the connection is congested so that they can (potentially) back off packet transmission (they could just ignore it, but it would be nice for them to be *able* to do something if they want).

Perhaps this might accomplish that:
- We keep a buffer of accepted datagrams.  Once we accept such a datagram then we'll send that datagram as soon as we can.
- We add a `private: std::atomic<bool> congested_;` along with a `public: bool is_congested() const { return congested_; }`
- Inside `flush_packets()` we do two things:
    - just after `if (n_packets > 0)` case, we set `congested_` to false
    - on any exits from the function that happen specifically because of congestion, we set `congested_ = true;` and drop all pending, queued datagrams.
- In the (application-facing) call to send a datagram, we check `is_congested()` and, if set, we immediately return a failure back to the user without queuing anything to the logic thread.  If not congested, we queue it.

Tom's suggestion:
- We add a configurable "max_pending_datagrams" settings on the connection (it could probably just be a public member, defaulting to 0), and then we make `send_datagram` accept if either `!congested_` *or* the current datagram queue is less than `max_pending_datagrams`.  Then instead of "drop all pending datagrams" on congestion we would only drop (from the *beginning* of the queue) as many as needed to lower the max pending datagrams to `max_pending_datagrams`.

include/quic/gnutls_crypto.hpp
tewinget: looks like this was marked resolved here but the relevant change wasn't committed.

include/quic/utils.hpp
```diff
@@ -37,28 +40,15 @@ extern "C"
...
 {
     inline auto log_cat = oxen::log::Cat("quic");
 
-    class Stream;
-
     using namespace std::literals;
     using namespace oxen::log::literals;
     using bstring = std::basic_string<std::byte>;
     using bstring_view = std::basic_string_view<std::byte>;
+    using buffer_que = std::deque<std::pair<bstring_view, std::shared_ptr<void>>>;
```
tewinget, line 52: `buffer_queue` or `buffer_deque` is probably better.

src/connection.cpp
```diff
@@ -64,7 +67,14 @@ namespace oxen::quic
         return 0;
     }
 
-    int recv_stream_data(
+    int on_recv_datagram(ngtcp2_conn* /* conn */, uint32_t flags, const uint8_t* data, size_t datalen, void* user_data)
+    {
+        log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+        return static_cast<Connection*>(user_data)->recv_datagram(
+                {reinterpret_cast<const std::byte*>(data), datalen}, flags & NGTCP2_STREAM_DATA_FLAG_FIN);
```
tewinget, line 75: Receive datagram has no `FIN` flag.  It seems to only have one flag, `0RTT`, which we don't support so don't care about.

The recv_datagram function below need not take a `bool fin` for the same reason.

src/endpoint.cpp
```diff
@@ -454,4 +484,9 @@ namespace oxen::quic
         return nullptr;
     }
 
+    bool Endpoint::in_event_loop() const
```
tewinget, line 491: I don't see this used anywhere (could always be github trolling).

tests/main.cpp
```diff
@@ -7,7 +7,7 @@ int main(int argc, char* argv[])
     Catch::Session session;
 
     using namespace Catch::Clara;
-    std::string log_level = "debug", log_file = "stderr";
+    std::string log_level = "trace", log_file = "stderr";
```
tewinget, line 12: idk if we want to default this to `trace` since it can easily be specified in the CLI.

include/quic/connection.hpp
```diff
@@ -288,6 +274,10 @@ namespace oxen::quic
 
         // holds a mapping of active streams
         std::map<int64_t, std::shared_ptr<Stream>> streams;
+        // datagram "pseudo-stream"
+        std::shared_ptr<DatagramIO> datagrams;
+        // stream "pseudo-stream" to finish off packets
```
tewinget: to finish off packets, yes, but also for ngtcp2 to create any packets it may need us to send like ACKs, retransmits, etc.

tewinget: Would be sufficient to just say "to represent ngtcp2 stream ID -1" though

include/quic/datagram.hpp
```diff
@@ -0,0 +1,95 @@
...
+        bool sent_fin() const override
+        {
+            log::debug(log_cat, "{} called", __PRETTY_FUNCTION__);
+            return false;
+        };
+        void set_fin(bool) override { log::debug(log_cat, "{} called", __PRETTY_FUNCTION__); };
+        size_t unsent() const override
+        {
+            log::debug(log_cat, "{} called", __PRETTY_FUNCTION__);
+            return std::numeric_limits<size_t>::min();
```
tewinget: Feels like this should have an actual value it could return.

include/quic/utils.hpp
```diff
@@ -108,6 +108,8 @@ namespace oxen::quic
     inline constexpr size_t MAX_LAZY_PMTUD_UDP_PAYLOAD = MAX_PMTUD_UDP_PAYLOAD;                // 1452
     inline constexpr size_t MAX_GREEDY_PMTUD_UDP_PAYLOAD = (MAX_LAZY_PMTUD_UDP_PAYLOAD << 1);  // 2904
 
+    inline constexpr size_t FLAG_WRITE_MORE = 0x01u;
```
tewinget: This should be set to ngtcp2's constant.  I think for now we have to assume the datagram and stream constants will always equal one another, but they *could* change, and nevertheless it's better to know where the number came from than to have a magic number.

src/connection.cpp
```diff
@@ -514,119 +395,172 @@ namespace oxen::quic
...
             {
                 auto& stream_ptr = it->second;
-                if (stream_ptr and not stream_ptr->sent_fin)
-                    strs.push_back(stream_ptr.get());
+                if (stream_ptr and not stream_ptr->_sent_fin)
+                    channels.push_back(stream_ptr.get());
             }
         }
 
+        size_t max_datagram_size = -1;
```
tewinget: It shouldn't matter, but this should probably default to 0 rather than -1, as `size_t` is unsigned.

src/connection.cpp
```diff
@@ -350,120 +395,172 @@ namespace oxen::quic
             for (auto it = mid; it != streams.end(); ++it)
             {
                 auto& stream_ptr = it->second;
-                if (stream_ptr and not stream_ptr->sent_fin)
-                    strs.push_back(stream_ptr.get());
+                if (stream_ptr and not stream_ptr->_sent_fin)
+                    channels.push_back(stream_ptr.get());
             }
```
jagerman, line 381: We should append the datagram channel right here rather than later, so that it also ends up randomly positioned in the initial channel list.

src/connection.cpp
```diff
@@ -514,119 +395,172 @@ namespace oxen::quic
...
+                        MAX_PMTUD_UDP_PAYLOAD,
+                        &ndatalen,
+                        flags,
+                        stream_id,
+                        bufs.data(),
+                        bufs.size(),
+                        ts);
+
+                log::trace(log_cat, "add_stream_data for stream {} returned [{},{}]", stream_id, nwrite, ndatalen);
+                // ndatalen "should" be >0 for streams and <0 for psuedo streams
```
tewinget: The docs seem sorta ambiguous, but it *looks* like for the `stream id == -1` case, ndatalen will be either -1 or 0, so not necessarily negative?

src/connection.cpp
```diff
@@ -514,119 +395,172 @@ namespace oxen::quic
...
+                    log::trace(log_cat, "Done writing: {} connection is congested");
+                    if (source->is_stream() && stream_id != -1)
+                        // we are congested, so clear all pending streams (aside from the -1
+                        // pseudo-stream at the end) so that our next call hits the -1 to finish off.
+                        channels.erase(channels.begin(), streams_end_it);
+                    continue;
+                }
+
+                // fatal error
+                if (nwrite < -500)
```
tewinget: should call `ngtcp2_err_is_fatal(nwrite)` or w/e that func is named.

src/connection.cpp
```diff
@@ -350,120 +395,172 @@ namespace oxen::quic
...
         ngtcp2_pkt_info pkt_info{};
         auto* buf_pos = reinterpret_cast<uint8_t*>(send_buffer.data());
         pkt_tx_timer_updater pkt_updater{*this, ts};
         size_t stream_packets = 0;
-        while (!strs.empty())
-        {
+        ngtcp2_ssize nwrite = 0;
+        ngtcp2_ssize ndatalen;
+        int datagram_accepted = 0;
+        bool write_successful = false;
```
jagerman: These variables should be inside the loop, just before used.  Otherwise their value can persist across loop iterations.

src/connection.cpp
```diff
@@ -350,120 +395,172 @@ namespace oxen::quic
...
+                        bufs.data(),
+                        1,
+                        ts);
+
+                log::debug(log_cat, "ngtcp2_conn_writev_datagram returned a value of {}", nwrite);
+
+                // both must be true to be successful
+                if (datagram_accepted != 0)
+                {
+                    // if the datagram was accepted at all, drop it from the buffer
```
jagerman:
```suggestion
                    // if the datagram was accepted, drop it from the buffer
```
The "at all" implies it could be partially accepted, but it cannot be (ngtcp2 guarantees this).

include/quic/datagram.hpp
```diff
@@ -0,0 +1,95 @@
...
+        virtual void send(bstring_view data, std::shared_ptr<void> keep_alive = nullptr) = 0;
+        virtual std::vector<ngtcp2_vec> pending() = 0;
+        virtual void append_buffer(bstring_view buffer, std::shared_ptr<void> keep_alive) = 0;
+        virtual int64_t stream_id() const = 0;
+        virtual bool is_closing() const = 0;
+        virtual bool sent_fin() const = 0;
+        virtual void set_fin(bool v) = 0;
+        virtual size_t unsent() const = 0;
+        virtual void wrote(size_t n) = 0;
+    };
```
jagerman, line 45: This class is useful internally for managing the buffers, but it feels like this internal implementation detail is leaking out to the libquic-using application and forcing complications on *outside* code to make *inside* code more maintainable.

From an application's point of view, I'm not sure what the value is in having Stream and DatagramIO inherit from a base type (and then giving me the base type): that just means the application has to untangle the object structure to get at the thing they care about (which will be a stream *or* a datagram, but rarely or perhaps never both).  And while we can duplicate most of Stream's public interface into IOChannel (as is done here) that is an awkward API for the caller when using datagrams because we are going through a base class that has a bunch of methods that are meaningless for a datagram channel.

If you were to use the base class *internally* but still have the user use Stream and DatagramIO, we would get more manageable code internally but also more manageable code externally, which seems like a win.

src/connection.cpp
jagerman: Perhaps rename to `packet_filled` or something along those lines?  (Unless I'm mistaken as to what this variable is meant to do)

include/quic/datagram.hpp
dr7ana: That's a really good point that I didn't consider. I guess I was trying to do 3 different things at the same time, and it "almost" worked
1) introduce the base IOChannel class and have datagram and stream inherit from them
2) combine the error handling for flushing packets into one efficient loop
3) have the user-provided callbacks interoperate with streams without issue

I'll untangle this, starting probably with point 2

src/connection.cpp
dr7ana: I was unsure, i never saw a ndatalen of 0 for the stream id of -1, and like you said the docs were ambiguous. ill look into it :+1:

src/connection.cpp
dr7ana: Thanks that was a dumb oversight

include/quic/datagram.hpp
dr7ana: For sure, I just needed a placeholder here in case it got accidentally called that wasn't in the realm of a realistic response

include/quic/gnutls_crypto.hpp
dr7ana: sorry I clicked resolve on these as we were chatting about them on mumble about it (and some of your later comments), it should be changed

include/quic/endpoint.hpp
```diff
@@ -53,7 +47,14 @@ namespace oxen::quic
         Endpoint(Endpoint&&) = delete;
         Endpoint& operator=(Endpoint&&) = delete;
 
-        explicit Endpoint(Network& n, const Address& listen_addr);
+        std::string local_addr() { return _local.to_string(); }
```
jagerman, line 50: I think this method was added for debug, but isn't really needed: if someone wants it as a string they have a few options: `ep.local().to_string()` or `"{}"_format(ep.local())`, or even just pass `ep.local()` if it's already going into a formattable string.

dr7ana: Sorry just saw this, will fix and rebase

include/quic/endpoint.hpp
```diff
@@ -90,12 +90,9 @@ namespace oxen::quic
             std::promise<std::shared_ptr<Connection>> p;
             auto f = p.get_future();
 
-            Path _path;
+            auto& _r = (_local.is_ipv6()) ? Address::map_ipv4_as_ipv6(remote) : remote;
```
tewinget: What happens if `remote` is also/already ipv6?

jagerman: This is making a reference to a temporary.

It would be cleaner to have the mapping function as a mutator of an Address rather than a static method making a new copy of it (because it seems unlikely that where this would be used would ever care about keeping the original).

jagerman: > What happens if `remote` is also/already ipv6?

Then you'll get an error as soon as you try to send anything.  I suppose that error could be detected earlier and throw, though.

dr7ana: Yeah I need to account for that error

@jagerman I made it into a static method because the Address passed to connect is a const reference, so I would have had to const cast it

dr7ana: fixed and added a test case

tewinget (approved):
Seems fine to me.

include/quic/endpoint.hpp
```diff
@@ -83,28 +85,29 @@ namespace oxen::quic
 
         // creates new outbound connection to remote; emplaces conn/interface pair in outbound map
         template <typename... Opt>
-        std::shared_ptr<connection_interface> connect(const Address& remote, Opt&&... opts)
+        std::shared_ptr<connection_interface> connect(Address& remote, Opt&&... opts)
```
jagerman:
```suggestion
        std::shared_ptr<connection_interface> connect(Address remote, Opt&&... opts)
```
so that it can be called with a const value (it means it gets passed by copy, but that's not a big deal).

include/quic/endpoint.hpp
```diff
@@ -83,28 +85,29 @@ namespace oxen::quic
...
         {
             std::promise<std::shared_ptr<Connection>> p;
             auto f = p.get_future();
 
-            net.call([&opts..., &p, path = Path{local, remote}, this]() mutable {
+            if (_local.is_ipv6() && not remote.is_ipv6())
```
jagerman:
```suggestion
            if (_local.is_ipv6() && !remote.is_ipv6())
```

include/quic/endpoint.hpp
dr7ana: will commit separately :+1:

include/quic/endpoint.hpp
dr7ana: same as above

jagerman (approved):

PR #18, opened 2023-07-04
feat: add Ninja to requirements

yougotwill:
Had to install Ninja (https://ninja-build.org/) manually on macOS so thought it was worth noting in the requirements.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -11,6 +11,7 @@ C++ library for multi-stream QUIC implementations.
 ### Requirements:
 
 - Cmake
+- Ninja
 - C++ 17 compiler (Clang, GCC, etc)
 - NGTCP2
 - GNUTLS


jagerman:
Ninja is definitely not required; it builds just fine with a regular makefile-based build.

yougotwill:
I'm not versed in make but from following the README instructions this is the initial error I got.

```
cmake .. -GNinja \
    -DCMAKE_C_COMPILER="clang" \
    -DCMAKE_CXX_COMPILER="clang++" \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_EXPORT_COMPILE_COMMANDS=1 \
    -DCMAKE_POLICY_DEFAULT_CMP0069=NEW
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
-- Configuring incomplete, errors occurred!
```

I'm guessing changing the CMAKE_MAKE_PROGRAM is what you are talking about?

tewinget:
yeah, it's the `-GNinja` bit in that `cmake` invocation which is telling `cmake` to use `ninja`.  We should definitely list it in the requirements if we're going to have that be the default way to build, but for now that's up in the air enough that it's meh.

dr7ana:
Thanks for pointing that out @yougotwill you're absolutely correct, it's something I added to the README at the very beginning and have not thought twice about it since. I will update the instructions on the next docs PR to address this and the other potential options that can be used with cmake; we'll also have a more robust build system for every platform. We (jason) just got CI up and running, so that will be updated after this round of refactors

PR #20, opened 2023-07-10
README updates for dependencies, cmake flags, etc.

jagerman:
- clean up `.gitignore` a bit
- various wording tweaks.
- add required versions to Requirements section
- remove ngtcp2 from requirements: we always build our own bundled version and so it isn't a requirement.
- add libevent to requirements
- add note about needing to clone submodules in addition to the main repo
- remove unneeded options from cmake invocation
- add `make` line (the current build instructions only configured the build, but didn't actually build it).
- add list of various cmake options someone might want to use after the basic version.
- add note about needing to generate certificates for test suite
- add note about speedtest-server/-client test binaries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-build/
-certs/
-.gdb_history
+/build*
 .cache/
 .vscode/
 .DS_Store
@@ -12,3 +10,4 @@ certs/
 rebuild.sh
 full_rebuild.sh
 notes.txt
+/compile_commands.json
--- a/readme.md
+++ b/readme.md
@@ -4,34 +4,54 @@ C++ library for multi-stream QUIC implementations.
 
 ## Overview
 
-"Bleeding-edge" is tracked in the dev branch, while the stable branch is intermittently updated as new features and capabilities are added.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,6 +1,4 @@
-build/
-certs/
```
dr7ana, line 2: I use this locally (certs)

jagerman: > I use this locally (certs)

Is it in the project root, or scattered around subdirs?

tewinget: can edit `.git/info/exclude` and it will act as a `.gitignore` for that clone/checkout of the repo.  I end up putting stuff like ctags files and helper scripts in there all the time.

dr7ana: @jagerman I stash a folder of certs for testing in `/tests/certs/*`, and my build script copies it into the right place for running the tests. It's not a huge loss if we drop it, tom's idea of adding it to the global gitignore is good :+1:

dr7ana (approved):

.gitignore
jagerman: Any reason not to just generate those on the fly in that script, using `utils/gen-certs.sh`?

PR #24, opened 2023-07-17
Datagram Splitting

dr7ana:
Datagram splitting PR, implementation details in commit messages

### DIFF
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -79,7 +79,6 @@ namespace oxen::quic
         virtual size_t get_max_datagram_size() const = 0;
         virtual bool datagrams_enabled() const = 0;
         virtual bool packet_splitting_enabled() const = 0;
-        virtual Splitting packet_splitting_policy() const = 0;
         virtual const ConnectionID& scid() const = 0;
 
         virtual ~connection_interface() = default;
@@ -142,7 +141,6 @@ namespace oxen::quic
         int get_max_streams() const override { return _max_streams; }
         bool datagrams_enabled() const override { return _datagrams_enabled; }
         bool packet_splitting_enabled() const override { return _packet_splitting; }
-        Splitting packet_splitting_policy() const override { return _policy; }
 
       private:
         // private Constructor (publicly construct via `make_conn` instead, so that we can properly
@@ -155,7 +153,6 @@ namespace oxen::quic
                 std::shared_ptr<IOContext> ctx,
                 ngtcp2_pkt_hd* hdr = nullptr);
 
-        ngtcp2_crypto_conn_ref cref;
         Endpoint& _endpoint;
         std::shared_ptr<IOContext> context;
         Direction dir;
@@ -166,73 +163,7 @@ namespace oxen::quic
         const int _max_streams{DEFAULT_MAX_BIDI_STREAMS};
         const bool _datagrams_enabled{false};
         const bool _packet_splitting{false};
-        const Splitting _policy{Splitting::NONE};
         std::atomic<bool> _congested{false};
-
-        /// Datagram Numbering:
-        /// Each datagram ID is incremented by four from the previous one, regardless of whether we are
-        /// splitting packets. The first 12 MSBs are the counter, and the 2 LSBs indicate if the packet
-        /// is split or not and which it is in the split (respectively). For example,
-        ///
-        ///     ID: 0bxxxx'xxxx'xxxx'xxzz
-        ///                            ^^
-        ///               split/nosplit|first or second packet
-        ///
-        /// Example - unsplit packets:
-        ///     Packet Number   |   Packet ID
-        ///         1           |       4           In the unsplit packet scheme, the dgram ID of each
-        ///         2           |       8           datagram satisfies the rule:
-        ///         3           |       12                          (ID % 4) == 0
-        ///         4           |       16          As a result, if a dgram ID is received that is a perfect
-        ///         5           |       20          multiple of 4, that endpoint is NOT splitting packets
-        ///
-        /// Example - split packets:
-        ///     Packet Number   |   Packet ID
-        ///         1                   6           In the split-packet scheme, the dgram ID of the first
-        ///         2                   7           of two datagrams satisfies the rule:
-        ///         3                   10                          (ID % 4) == 2
-        ///         4                   11          The second of the two datagrams satisfies the rule:
-        ///         5                   14                          (ID % 4) == 3
-        ///         6                   15          As a result, a packet-splitting endpoint should never send
-        ///                                         or receive a datagram whose ID is a perfect multiple of 4
-        ///
-        uint64_t _increment = 4;
-        uint64_t _last_dgram_id{4};
-        int _last_cleared{-1};
-
-        // Increments _last_dgram_id by _increment, returns next ID
-        uint64_t next_dgram_id();
-
-        // Holds datagrams on deck to be sent
-        buffer_que unsent_datagrams;
-
-        /// Holds received datagrams in a rotating "tetris" ring-buffer arrangement of split, unmatched packets.
-        /// When a datagram with ID N is recieved, we store it as:
-        ///
-        ///         tetris_buffer[i][j]
-        /// where,
-        ///         i = (N % 4096) / 1024
-        ///         j = N % 1024
-        ///
-        /// When it comes to clearing the buffers, the last cleared row is stored in Connection::_last_cleared.
-        /// The next row to clear is found as:
-        ///
-        ///         to_clear = (i + 2) % 4;
-        ///         if (to_clear == (last_cleared+2)%4)
-        ///         {
-        ///             clear(to_clear)
-        ///             last_cleared = to_clear
-        ///         }
-        ///
-        /// In full, given 'last_cleared' and a current tetris_buffer index 'i', we find 'to_clear' to be:
-        ///     last_cleared  |  i  |  to_clear
-        /// (init) -1                    1
-        ///         0                    2
-        ///         1                    3
-        ///         2                    0
-        ///         3                    1
-        ///
-        std::array<std::array<Packet*, 1024>, 4> tetris_buffer;
 
         struct connection_deleter
         {
@@ -281,7 +212,11 @@ namespace oxen::quic
 
         int init(ngtcp2_settings& settings, ngtcp2_transport_params& params, ngtcp2_callbacks& callbacks);
 
+        io_result read_packet(const Packet& pkt);
+
       public:
+        // public to be called by endpoint handing this connection a packet
+        void handle_conn_packet(const Packet& pkt);
         // these are public so ngtcp2 can access them from callbacks
         int stream_opened(int64_t id);
         int stream_ack(int64_t id, size_t size);
@@ -290,7 +225,6 @@ namespace oxen::quic
         void check_pending_streams(int available);
         int recv_datagram(bstring_view data, bool fin);
         int ack_datagram(uint64_t dgram_id);
-        int lost_datagram(uint64_t dgram_id);
 
         // Implicit conversion of Connection to the underlying ngtcp2_conn* (so that you can pass a
         // Connection directly to ngtcp2 functions taking a ngtcp2_conn* argument).
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -45,18 +45,15 @@ namespace oxen::quic
     class debug_interface
     {
       public:
-        std::atomic<bool> enable_datagram_drop_test;
-        std::atomic<int> test_drop_counter;
-        std::atomic<bool> enable_datagram_flip_flop_test;
-        std::atomic<int> test_flip_flop_counter;
+        std::atomic<bool> datagram_drop_enabled;
+        std::atomic<int> datagram_drop_counter;
+        std::atomic<bool> datagram_flip_flop_enabled;
+        std::atomic<int> datagram_flip_flip_counter;
     };
 
 #endif
 
     class connection_interface
-#ifndef NDEBUG
-            : public debug_interface
-#endif
     {
       public:
         virtual std::shared_ptr<Stream> get_new_stream(
@@ -96,12 +93,21 @@ namespace oxen::quic
         virtual bool packet_splitting_enabled() const = 0;
         virtual const ConnectionID& scid() const = 0;
 
-        // public debug functions: remove this with testfixture friend classes
-        virtual int datagrams_stored() = 0;
-        virtual int last_cleared() = 0;
-        virtual int datagram_bufsize() = 0;
+        // WIP functions: these are meant to expose specific aspects of the internal state of connection
+        // and the datagram IO object for debugging and application (user) utilization.
+        //
+        //  datagrams_stored: returns the number of partial datagrams held waiting for their counterpart
+        //  last_cleared: returns the index of the last cleared bucket in the recv_buffer
+        //  datagram_bufsize: returns the total number of datagrams that the recv_buffer can hold
+        virtual int datagrams_stored() const = 0;
+        virtual int last_cleared() const = 0;
+        virtual int datagram_bufsize() const = 0;
 
         virtual ~connection_interface() = default;
+
+#ifndef NDEBUG
+        debug_interface test_suite;
+#endif
     };
 
     class Connection : public connection_interface, public std::enable_shared_from_this<Connection>
@@ -161,9 +167,9 @@ namespace oxen::quic
         bool packet_splitting_enabled() const override { return _packet_splitting; }
 
         // public debug functions; to be removed with friend test fixture class
-        int datagrams_stored() override;  // TOFIX: this one
-        int last_cleared() override;
-        int datagram_bufsize() override;
+        int datagrams_stored() const override;  // TOFIX: this one
+        int last_cleared() const override;
+        int datagram_bufsize() const override;
 
       private:
         // private Constructor (publicly construct via `make_conn` instead, so that we can properly
@@ -264,13 +270,6 @@ namespace oxen::quic
 
         // returns number of currently pending streams for use in test cases
         size_t num_pending() const { return pending_streams.size(); }
-
-#ifndef NDEBUG
-        // static std::atomic<bool> enable_datagram_drop_test;
-        // static std::atomic<int> test_drop_counter;
-        // static std::atomic<bool> enable_datagram_flip_flop_test;
-        // static std::atomic<int> test_flip_flop_counter;
-#endif
     };
 
     extern "C"
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -41,7 +41,22 @@ namespace oxen::quic
     template <>
     constexpr inline bool IsToStringFormattable<ConnectionID> = true;
 
+#ifndef NDEBUG
+    class debug_interface
+    {
+      public:
+        std::atomic<bool> enable_datagram_drop_test;
+        std::atomic<int> test_drop_counter;
+        std::atomic<bool> enable_datagram_flip_flop_test;
+        std::atomic<int> test_flip_flop_counter;
+    };
+
+#endif
+
     class connection_interface
+#ifndef NDEBUG
+            : public debug_interface
+#endif
     {
       public:
         virtual std::shared_ptr<Stream> get_new_stream(
@@ -249,6 +264,13 @@ namespace oxen::quic
 
         // returns number of currently pending streams for use in test cases
         size_t num_pending() const { return pending_streams.size(); }
+
+#ifndef NDEBUG
+        // static std::atomic<bool> enable_datagram_drop_test;
+        // static std::atomic<int> test_drop_counter;
+        // static std::atomic<bool> enable_datagram_flip_flop_test;
+        // static std::atomic<int> test_flip_flop_counter;
+#endif
     };
 
     extern "C"
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -71,7 +71,7 @@ namespace oxen::quic
             send_datagram(view, std::move(keep_alive));
         }
 
-        // TOFIX: We can't template virtual functions; do this a better way if it ends up working
+        // note: We can't template virtual functions; do this a better way if it ends up working
         virtual void send_datagram(bstring_view data, std::shared_ptr<void> keep_alive = nullptr) = 0;
 
         virtual int get_max_streams() const = 0;
@@ -80,6 +80,11 @@ namespace oxen::quic
         virtual bool datagrams_enabled() const = 0;
         virtual bool packet_splitting_enabled() const = 0;
         virtual const ConnectionID& scid() const = 0;
+
+        // public debug functions: remove this with testfixture friend classes
+        virtual int datagrams_stored() = 0;
+        virtual int last_cleared() = 0;
+        virtual int datagram_bufsize() = 0;
 
         virtual ~connection_interface() = default;
     };
@@ -98,9 +103,7 @@ namespace oxen::quic
         //      scid: local ("primary") CID used for this connection (random for outgoing)
         //		dcid: remote CID used for this connection
         //      path: network path used to reach remote client
-        //		creds: relevant tls information per connection
-        //		u_config: user configuration values passed in struct
-        //      dir: enum specifying configuration detailts for client vs. server
+        //      ctx: IO session dedicated for this connection context
         //		hdr: optional parameter to pass to ngtcp2 for server specific details
         static std::shared_ptr<Connection> make_conn(
                 Endpoint& ep,
@@ -141,6 +144,11 @@ namespace oxen::quic
         int get_max_streams() const override { return _max_streams; }
         bool datagrams_enabled() const override { return _datagrams_enabled; }
         bool packet_splitting_enabled() const override { return _packet_splitting; }
+
+        // public debug functions; to be removed with friend test fixture class
+        int datagrams_stored() override;  // TOFIX: this one
+        int last_cleared() override;
+        int datagram_bufsize() override;
 
       private:
         // private Constructor (publicly construct via `make_conn` instead, so that we can properly
@@ -203,7 +211,7 @@ namespace oxen::quic
         // holds a mapping of active streams
         std::map<int64_t, std::shared_ptr<Stream>> streams;
         // datagram "pseudo-stream"
-        std::shared_ptr<DatagramIO> datagrams;
+        std::unique_ptr<DatagramIO> datagrams;
         // "pseudo-stream" to represent ngtcp2 stream ID -1
         std::shared_ptr<Stream> pseudo_stream;
         // holds queue of pending streams not yet ready to broadcast
--- a/include/quic/datagram.hpp
+++ b/include/quic/datagram.hpp
@@ -5,8 +5,11 @@
 
 namespace oxen::quic
 {
-    extern std::atomic<bool> enable_test_features;
+#ifndef NDEBUG
+    extern std::atomic<bool> enable_datagram_drop_test;
+    extern std::atomic<bool> enable_datagram_flip_flop_test;
     extern std::atomic<int> test_counter;
+#endif
 
     class IOChannel;
     class Connection;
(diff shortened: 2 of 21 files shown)


tests/007-datagrams.cpp
```diff
@@ -160,87 +151,115 @@ namespace oxen::quic::test
...
         SECTION("Simple datagram transmission")
         {
             Network test_net{};
-            auto msg = "hello from the other siiiii-iiiiide"_bsv;
+            std::string data{};
+            char v = 0;
 
-            std::promise<bool> tls_promise, data_promise;
-            std::future<bool> tls_future = tls_promise.get_future(), data_future = data_promise.get_future();
+            while (data.size() < 2000)
```
tewinget: This number may as well reflect max datagram size.

tewinget:
Since we're only splitting in half, there should be a test which attempts to send a packet which won't fit in 2 datagrams and requires that `send_datagram` throws.

tewinget:
We'll also want tests to make sure the rotating buffer behavior is correct.  That should be interesting to design positive and negative tests for, and of course then implement.

tests/007-datagrams.cpp
dr7ana: good thought, will reflect in the updated 007 test when i add the rotating buffer test

tests/007-datagrams.cpp
```diff
@@ -279,15 +298,317 @@ namespace oxen::quic::test
...
+            SKIP("Rotating buffer testing not enabled for this test iteration!");
+
+        SECTION("Simple oversized datagram transmission - Clear first row")
+        {
+            log::trace(log_cat, "Beginning the unit test from hell");
+            Network test_net{};
+
+            std::atomic<int> index{0};
+            std::atomic<int> data_counter{0};
+            size_t bufsize = 256, n = bufsize / 2 + 1;
```
tewinget: not a huge fan of multiple variable declarations on the same line.  also since `n` is used in several places and is not like a simple counter,  it could use a more descriptive name.

include/quic/opt.hpp
```diff
@@ -43,10 +43,16 @@ namespace oxen::quic::opt
     {
         bool split_packets = false;
         Splitting mode = Splitting::NONE;
+        int bufsize = 4096;
```
tewinget, line 50: should probably be a `size_t` or some other unsigned type.

include/quic/datagram.hpp
```diff
@@ -1,16 +1,71 @@
...
     // IO callbacks
-    using dgram_data_callback = std::function<void(bstring_view)>;
+    using dgram_data_callback = std::function<void(bstring)>;
+
+    using dgram_buffer = std::deque<std::pair<uint16_t, std::pair<bstring_view, std::shared_ptr<void>>>>;
+
+    struct prepared_datagram
+    {
+        uint64_t id;                  // internal ID for ngtcp2
+        std::array<uint8_t, 2> dgid;  // optional transmitted ID buffer (for packet splitting)
```
jagerman: Shouldn't this be just a uint16_t?

jagerman: Wait, nevermind, I see how this is used.

tests/007-datagrams.cpp
```diff
@@ -279,15 +298,317 @@ namespace oxen::quic::test
...
+                conn_interface->send_datagram(std::basic_string_view<uint8_t>{good_msg});
+
+            for (auto& f : data_futures)
+                REQUIRE(f.get());
+
+            REQUIRE(data_counter == int(n));
+
+            auto server_ci = server_endpoint->get_all_conns(Direction::INBOUND).front();
+
+            REQUIRE(server_ci->last_cleared() == 0);
```
tewinget, line 415: This confirms that `last_cleared` was set to the expected value, but it doesn't confirm that the bucket was actually cleared.  In order to do that, we would need to have a way to make the connection send one half of a packet but not the other half, and then we'd need to send enough packets to loop back around do that packet number and make sure the other side of the connection gets the correct new packet rather than some frankensteined packet.

dr7ana: keep scrolling, theres a test case that does exactly that

include/quic/datagram.hpp
```diff
@@ -1,16 +1,71 @@
...
+        const ngtcp2_vec* data() const { return bufs.data(); }
+        size_t size() const { return bufs_len; }
+    };
+
+    struct received_datagram
+    {
+        uint16_t id{0};
+        bstring data{};
+        // -1 = first half, 1 = second half
+        int part{0};
```
jagerman: Any reason not just to use a bool (maybe `first_part`) for this?

Also, whether bool or not, this would be slightly better placed between `id` and `data` to make the struct layout *slightly* smaller: `bstring` most likely uses size_t-alignment (i.e. 8-byte alignment), so if you do it this way you have a struct containing data like this: `[u16][6 bytes padding][string, probably 24 bytes][int][4 bytes padding]`, whereas if you rearrange the smaller types together you will get: `[u16][2 padding][int][24-byte-string]` and make the struct 8 bytes smaller.  Not a huge deal, but also easy to change.

dr7ana: I used an int so I can reset it to 0 when the entry is cleared

Good point about the alignment, will fix it

include/quic/datagram.hpp
```diff
@@ -1,16 +1,71 @@
 #pragma once
 
+#include "udp.hpp"
 #include "utils.hpp"
 
 namespace oxen::quic
 {
+    extern std::atomic<bool> enable_test_features;
```
jagerman: Perhaps rename to `debug_drop_datagrams`?

include/quic/datagram.hpp
```diff
@@ -1,16 +1,71 @@
...
+    extern std::atomic<int> test_counter;
```
jagerman: These two should also be inside `#ifndef NDEBUG`

tests/007-datagrams.cpp
```diff
@@ -279,15 +298,317 @@ namespace oxen::quic::test
...
 
             test_net.close();
         };
     };
+
+    TEST_CASE("007 - Datagram support: Rotating Buffer, Induced Loss", "[007][datagrams][execute][split][rotating][loss]")
+    {
+        if (disable_rotating_buffer)
+            SKIP("Rotating buffer testing not enabled for this test iteration!");
+
```
jagerman:
```suggestion

#ifdef NDEBUG
        SKIP("Induced loss test requires a debug build");
#else
```

jagerman: and `#endif` down at the bottom of the section

dr7ana: added!

src/connection.cpp
```diff
@@ -740,14 +865,41 @@ namespace oxen::quic
...
+    }
+
     int Connection::recv_datagram(bstring_view data, bool fin)
     {
         log::trace(log_cat, "Connection (CID: {}) received datagram: {}", _source_cid, buffer_printer{data});
 
-        // append received datagram to tetris_buffer if packet_splitting is enabled
-        if (_packet_splitting)
+        std::optional<bstring> maybe_data;
+        uint16_t dgid = oxenc::load_big_to_host<uint16_t>(data.data());
```
jagerman: When packet splitting is completely disabled on the connection we ~~shouldn't be~~ *aren't* sending a dgid, so need to not try to read one here.

src/connection.cpp
```diff
@@ -758,7 +910,7 @@ namespace oxen::quic
 
             try
             {
-                datagrams->dgram_data_cb(data);
+                datagrams->dgram_data_cb((maybe_data ? *maybe_data : bstring{data.begin(), data.end()}));
```
jagerman:
```suggestion
                datagrams->dgram_data_cb((maybe_data ? std::move(*maybe_data) : bstring{data.begin(), data.end()}));
```

src/connection.cpp
dr7ana: fixed

src/connection.cpp
dr7ana: fixed!

src/connection.cpp
```diff
@@ -740,14 +865,41 @@ namespace oxen::quic
...
-            // TODO: should this logic go earlier in the process inside handle_packet?
+            try
+            {
+                // send received datagram to tetris_buffer if packet_splitting is enabled
+                maybe_data = datagrams->to_buffer(data, dgid);
+            }
+            catch (std::exception& e)
+            {
+                log::error(log_cat, "Exception: {}", e.what());
+                return -1;
```
jagerman: Should this be NGTCP2_ERR_CALLBACK_FAILURE?

src/datagram.cpp
```diff
@@ -3,73 +3,257 @@
...
+
+        return std::nullopt;
+    }
+
+    std::optional<bstring> rotating_buffer::receive(bstring_view data, uint16_t dgid)
+    {
+        log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+
+        {
+            std::lock_guard<std::mutex> buffer_lock(m);
```
jagerman: I *think* we can drop the mutex and atomics in here, if we ensure that we only ever touch them from the main thread.  (For the test suite I think that would mean we need to add some indirection that goes through a `call` rather than accessing things directly, but that's probably fine for the test suite).

src/datagram.cpp
```diff
@@ -3,73 +3,257 @@
...
+        return d;
+    }
+
+    std::optional<bstring> DatagramIO::to_buffer(bstring_view data, uint16_t dgid)
+    {
+        log::trace(log_cat, "DatagramIO handed datagram with endian swapped ID: {}", dgid);
 
-            temp.base = const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(b.first.data()));
-            temp.len = b.first.size();
+        auto r = recv_buffer.receive(data, dgid);
```
jagerman: I think we can just `return recv_buffer.receive(data, dgid);` here?

src/datagram.cpp
```diff
@@ -3,73 +3,257 @@
...
+
+                log::trace(log_cat, "Pairing datagram (ID: {}) with counterpart at buffer pos [{},{}]", dgid, row, col);
+
+                bstring out{};
+                out.reserve(data.size() + b.data.size());
+
+                bool order = (b.part < 0);  // if true, we have the first part already stored
+                // must use substring or we will end up with two datagram ID's at the beginning and middle
+                out.append((order ? b.data.substr(2) : data.substr(2)));
+                out.append((order ? data.substr(2) : b.data.substr(2)));
```
jagerman: Unfortunately, string.substr allocates and copies into a brand new string.  You could avoid it by wrapping in a string_view first, like this:
```suggestion
                out.append(bstring_view{order ? b.data : data}.substr(2));
                out.append(bstring_view{order ? data : b.data}.substr(2));
```

src/datagram.cpp
```diff
@@ -3,73 +3,257 @@
...
+                else
+                {
+                    log::debug(log_cat, "enable_test_features is false, skipping optional logic");
+                }
+#endif
+
+                log::trace(log_cat, "Pairing datagram (ID: {}) with counterpart at buffer pos [{},{}]", dgid, row, col);
+
+                bstring out{};
+                out.reserve(data.size() + b.data.size());
```
jagerman: I think that we could potentially avoid another string allocation here by reusing either `b.data` or `data`.  Maybe something like:

```C++
bstring out;
if (order) {
    b.data.append(data);
    out = std::move(b.data);
} else {
    data.append(b.data);
    out = std::move(data);
}
```

Then we could *also* reserve space when we first set up the partial `.data` element so that this append won't need to reallocate.

src/datagram.cpp
jagerman: Perhaps better would be to do the substring earlier, so that we don't keep the dgid in the data bstring at all

src/datagram.cpp
```diff
@@ -3,73 +3,257 @@
...
+                out.append((order ? data.substr(2) : b.data.substr(2)));
+
+                b.clear_entry();
+
+                return out;
+            }
+
+            log::trace(log_cat, "Storing datagram (ID: {}) at buffer pos [{},{}]", dgid, row, col);
+
+            b = received_datagram{dgid, data};
```
jagerman: ... that is, just before this line we can do `data.remove_prefix(2);` to discard the dgid from the data we store.

src/datagram.cpp
jagerman: No, wait, the second version here is wrong.

jagerman: v2:
```C++
if (order)
    b.data.append(data);
else {
    b.data.reserve(b.data.size() + data.size());
    // We have the second half and need to prepend the first half, so move it to the end:
    std::memmove(b.data.data() + data.size(), b.data.data(), b.data.size());
    // Now copy the first half into place:
    std::memcpy(b.data.data(), data.data(), data.size());
}
bstring out{std::move(b.data)};
```
which should need no new allocation *if* we pre-reserved the needed amount when creating the data.

include/quic/datagram.hpp
```diff
@@ -1,16 +1,71 @@
...
+
+    struct received_datagram
+    {
+        uint16_t id{0};
+        bstring data{};
+        // -1 = first half, 1 = second half
+        int part{0};
+
+        received_datagram() = default;
+        explicit received_datagram(uint16_t dgid, bstring_view d) : id{dgid}, data{d} { part = (dgid % 4 == 2) ? -1 : 1; };
```
jagerman: We know that *at most* the assembled datagram will be `d.size() + MAX_DATAGRAM_SIZE` (this is probably the wrong name for that constant), so we can reserve it with:
```suggestion
        explicit received_datagram(uint16_t dgid, bstring_view d) : id{dgid}, part{(dgid % 4 == 2) ? -1 : 1} {
            data.reserve(d.size() + MAX_DATAGRAM_SIZE);
            data.append(d);
        }
```
and then we fill out this buffer with the remaining packet when it arrives without having to copy and allocate a new buffer (left a comment over there about this).

src/datagram.cpp
jagerman: Also those three lines have a lot of Data.  Lore will be upset.

src/connection.cpp
dr7ana: that's a good question... on one hand it doesn't matter because any non-zero return value buffs out to `NGTCP2_ERR_CALLBACK_FAILURE` [upon return](https://github.com/ngtcp2/ngtcp2/blob/d80f9b3d709da79cdbd2b94acfae3073810de80d/lib/ngtcp2_conn.c#L589-L593) so we might as well return that... on the other hand that specific line isn't a failure in the callback, but in the buffer.

maybe we could extend std exception and make our own that we throw and handle internally?

src/datagram.cpp
dr7ana: yeah that's a good point, didn't think about that! i will do what your later comment suggests and drop the prefix earlier

src/datagram.cpp
dr7ana: or even earlier in connection?

include/quic/datagram.hpp
dr7ana: awesome

src/datagram.cpp
dr7ana: quick note about the memmoves: when we already have the first half of the datagram, the incoming smaller half is overwriting the big half. this is with the new storage structures, so ill adjust and let you know

jagerman: The `reserve` should be a `resize` instead.

src/datagram.cpp
```diff
@@ -3,73 +3,275 @@
...
+                    (b.part < 0 ? "first"s : "second"s),
+                    row,
+                    col);
+
+            if (b.part < 0)  // if true, we have the first part already stored
+                b.data.append(data);
+            else
+            {
+                b.data.reserve(b.data.size() + data.size());
+                b.data.insert(0, data.data());
```
jagerman: Ah yeah, that's cleaner than my version.

include/quic/connection.hpp
```diff
@@ -307,6 +264,13 @@ namespace oxen::quic
 
         // returns number of currently pending streams for use in test cases
         size_t num_pending() const { return pending_streams.size(); }
+
+#ifndef NDEBUG
```
tewinget: supplanted by `debug_interface`

include/quic/connection.hpp
```diff
@@ -41,7 +41,22 @@ namespace oxen::quic
     template <>
     constexpr inline bool IsToStringFormattable<ConnectionID> = true;
 
+#ifndef NDEBUG
+    class debug_interface
+    {
+      public:
+        std::atomic<bool> enable_datagram_drop_test;
```
tewinget: reads like a verb

include/quic/connection.hpp
```diff
@@ -71,17 +86,21 @@ namespace oxen::quic
             send_datagram(view, std::move(keep_alive));
         }
 
-        // TOFIX: We can't template virtual functions; do this a better way if it ends up working
+        // note: We can't template virtual functions; do this a better way if it ends up working
```
jagerman, line 86: You can make this work by having a `protected` base pure virtual method (maybe called `send_impl` or something like that), and then templatizing as much as you want here in `connection_interface` where each templated version recasts its arguments and passes them off to `send_impl` (so that an implementing class need only implement/override the single `send_impl` method).

include/quic/connection.hpp
```diff
@@ -142,7 +159,11 @@ namespace oxen::quic
         int get_max_streams() const override { return _max_streams; }
         bool datagrams_enabled() const override { return _datagrams_enabled; }
         bool packet_splitting_enabled() const override { return _packet_splitting; }
-        Splitting packet_splitting_policy() const override { return _policy; }
+
+        // public debug functions; to be removed with friend test fixture class
+        int datagrams_stored() override;  // TOFIX: this one
+        int last_cleared() override;
+        int datagram_bufsize() override;
```
jagerman: ~~To remove?~~

include/quic/stream.hpp
```diff
@@ -219,6 +223,12 @@ namespace oxen::quic
             }
         };
 
+        prepared_datagram pending_datagram(std::atomic<bool>&) override
```
tewinget: This should probably throw?

include/quic/datagram.hpp
```diff
@@ -1,16 +1,26 @@
...
 
 namespace oxen::quic
 {
-    class IOChannel;
+#ifndef NDEBUG
+    // extern std::atomic<bool> enable_datagram_drop_test;
+    // extern std::atomic<int> test_drop_counter;
+    // extern std::atomic<bool> enable_datagram_flip_flop_test;
+    // extern std::atomic<int> test_flip_flop_counter;
+#endif
```
jagerman: Delete

include/quic/datagram.hpp
```diff
@@ -32,16 +41,16 @@ namespace oxen::quic
         virtual bool is_stream() = 0;
         virtual bool is_empty() const = 0;
         virtual std::shared_ptr<Stream> get_stream() = 0;
-        virtual void send(bstring_view data, std::shared_ptr<void> keep_alive = nullptr) = 0;
+        virtual void send(bstring_view, std::shared_ptr<void> keep_alive = nullptr) = 0;
         virtual std::vector<ngtcp2_vec> pending() = 0;
-        virtual void append_buffer(bstring_view buffer, std::shared_ptr<void> keep_alive) = 0;
+        virtual prepared_datagram pending_datagram(std::atomic<bool>&) = 0;
```
jagerman: Not clear why this needs to be atomic

src/connection.cpp
```diff
@@ -467,9 +583,15 @@ namespace oxen::quic
                         ts);
 
                 log::trace(log_cat, "add_stream_data for stream {} returned [{},{}]", stream_id, nwrite, ndatalen);
+
+                // if we just loaded stream frames into the packet, we should not reverse the order of the next split
+                // datagram
+                // prefer_big_first = false;
```
tewinget: if this is dead code, should be removed

src/connection.cpp
```diff
@@ -417,6 +531,8 @@ namespace oxen::quic
         pkt_tx_timer_updater pkt_updater{*this, ts};
         size_t stream_packets = 0;
 
+        std::atomic<bool> prefer_big_first{true};
```
tewinget: shouldn't need to be atomic

src/connection.cpp
```diff
@@ -478,15 +600,19 @@ namespace oxen::quic
...
                         ts);
 
                 log::debug(log_cat, "ngtcp2_conn_writev_datagram returned a value of {}", nwrite);
 
                 if (datagram_accepted != 0)
-                    datagrams->user_buffers.pop_front();
+                {
+                    log::trace(log_cat, "ngtcp2 accepted datagram ID: {} for transmission", dgram.id);
+                    datagrams->send_buffer.drop_front(prefer_big_first);
+                    // datagrams->send_buffer.erase(dgram.target);
```
tewinget: dead code

include/quic/connection.hpp
jagerman: Never mind, these are (potentially) useful for an application that maybe wants to worry about how packets are arriving.

Perhaps add a one-line comment above each to give a quick description of what they are?

include/quic/messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+        // -1 = payload, 1 = addendum
+        int part{0};
+        bstring data{};
+
+        received_datagram() = default;
+        explicit received_datagram(uint16_t dgid, bstring_view d) : id{dgid}, part{(dgid % 4 == 2) ? -1 : 1}
+        {
+            data.reserve(d.size() + MAX_PMTUD_UDP_PAYLOAD);
+            data.append(d);
+        };
```
jagerman:
```suggestion
        }
```

include/quic/messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+        void clear_entry()
+        {
+            id = 0;
+            part = 0;
+            data.clear();
+        };
```
jagerman:
```suggestion
        }
```

src/connection.cpp
jagerman: ~~Should this be uncommented?~~

Dead code.

include/quic/connection.hpp
```diff
@@ -71,17 +86,21 @@ namespace oxen::quic
...
         virtual size_t get_max_datagram_size() const = 0;
         virtual bool datagrams_enabled() const = 0;
         virtual bool packet_splitting_enabled() const = 0;
-        virtual Splitting packet_splitting_policy() const = 0;
         virtual const ConnectionID& scid() const = 0;
 
+        // public debug functions: remove this with testfixture friend classes
+        virtual int datagrams_stored() = 0;
+        virtual int last_cleared() = 0;
+        virtual int datagram_bufsize() = 0;
```
jagerman: I think these can all be `const`

include/quic/datagram.hpp
```diff
@@ -84,14 +172,30 @@ namespace oxen::quic
...
+/*
+
+Revolving send buffer:
+    - both datagram halves have the same right shifted index
+
+Receive buffer:
+    - more rows can increase tolerance in misordering
+    - too many short ones leads to pinstriping
+
+*/
```
jagerman: Leftover notes?

include/quic/datagram.hpp
```diff
@@ -84,14 +172,30 @@ namespace oxen::quic
...
+                sum += entry.size();
             return sum;
         };
-        bool has_unsent() const override { return is_empty(); }
+        bool has_unsent() const override { return not is_empty(); }
         void wrote(size_t) override { log::debug(log_cat, "{} called", __PRETTY_FUNCTION__); };
+        std::vector<ngtcp2_vec> pending() override
+        {
+            log::warning(log_cat, "{} called", __PRETTY_FUNCTION__);
+            return std::vector<ngtcp2_vec>{};
```
jagerman: Purely a style thing (i.e. feel completely free to ignore), but this can be written as `return {};` if you don't feel like repeating the return type again.

include/quic/stream.hpp
```diff
@@ -219,6 +223,12 @@ namespace oxen::quic
...
+        {
+            log::warning(log_cat, "{} called", __PRETTY_FUNCTION__);
+            return prepared_datagram{};
+        };
```
jagerman:
```suggestion
        }
```

src/connection.cpp
```diff
@@ -191,15 +217,106 @@ namespace oxen::quic
...
 
-    void Connection::packet_io_ready()
+    void Connection::handle_conn_packet(const Packet& pkt)
     {
-        event_active(packet_io_trigger.get(), 0, 0);
+        if (auto rv = ngtcp2_conn_in_closing_period(*this); rv != 0)
+        {
+            log::trace(log_cat, "Note: CID-{} in closing period; signaling endpoint to delete connection", scid());
+
+            _endpoint.call([&]() {
```
jagerman: Another completely ignorable style point: for a lambda with no arguments you can leave out the `()` entirely, i.e.
```suggestion
            _endpoint.call([&] {
```

src/connection.cpp
```diff
@@ -294,8 +411,6 @@ namespace oxen::quic
         }
     };
 
-    int64_t sent_counter = 0;
-
```
jagerman, line 298: Whoops, my bad here leaving that in in the udp refactor.

src/connection.cpp
jagerman: Since only `flush_packets` touches this (and flush_packets will only ever be called inside the main thread) this ought to be safe as a plain `bool`

src/connection.cpp
jagerman: Dead code

src/datagram.cpp
```diff
@@ -3,73 +3,112 @@
 #include "connection.hpp"
 #include "endpoint.hpp"
 
+#ifndef NDEBUG
+// std::atomic<bool> oxen::quic::enable_datagram_drop_test;
+// std::atomic<bool> oxen::quic::enable_datagram_flip_flop_test;
+// std::atomic<int> oxen::quic::test_drop_counter;
+// std::atomic<int> oxen::quic::test_flip_flop_counter;
+#endif
```
jagerman: dead code

src/datagram.cpp
```diff
@@ -3,73 +3,112 @@
...
+        //         {
+        //             log::debug(log_cat, "enable_datagram_flip_flop_test is true, bypassing buffer");
+        //             conn.test_flip_flop_counter += 1;
+        //             log::debug(log_cat, "test counter: {}", conn.test_flip_flop_counter.load());
+        //             return std::nullopt;
+        //         }
+        //         else
+        //         {
+        //             log::debug(log_cat, "enable_datagram_flip_flop_test is false, skipping optional logic");
+        //         }
```
jagerman: dead code

include/quic/messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+        std::optional<bstring> receive(bstring_view data, uint16_t dgid);
+        void clear_row(int index);
+        int datagrams_stored();
+    };
+
+    struct buffer_que
+    {
+        std::deque<datagram_storage> buf{};
+        size_t quantity{0};
```
tewinget: appears unused

include/quic/messages.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+        bool empty() const { return buf.empty(); }
+        size_t size() const { return quantity; }
```
tewinget: should return `buf.size()` ?

src/connection.cpp
jagerman: Here it might be cleaner/tighter to capture `[this]` rather than `[&]` as it looks like the current object is the only thing we care about capturing inside the lambda.

src/connection.cpp
```diff
@@ -191,15 +217,106 @@ namespace oxen::quic
...
+            });
+            return;
+        }
+
+        if (is_draining())
+        {
+            log::debug(log_cat, "Error: connection is already draining; dropping");
+        }
+
+        // TODO: if read packet gives us failure, should we close?
```
tewinget: This appears to be covered within `read_packet`

src/connection.cpp
```diff
@@ -191,15 +217,106 @@ namespace oxen::quic
...
+        auto rv = ngtcp2_conn_read_pkt(*this, pkt.path, &pkt.pkt_info, u8data(pkt.data), pkt.data.size(), ts);
+
+        switch (rv)
+        {
+            case 0:
+                packet_io_ready();
+                break;
+            case NGTCP2_ERR_DRAINING:
+                log::trace(log_cat, "Note: CID-{} is draining; signaling endpoint to drain connection", scid());
+                _endpoint.call([&]() {
```
jagerman: These (this one and the various ones that follow) could also be `[this]` captures

include/quic/connection.hpp
dr7ana: resolving this after our mumble chat

include/quic/connection.hpp
dr7ana: just do it

include/quic/stream.hpp
dr7ana: good idea

include/quic/connection.hpp
dr7ana: you got it

include/quic/datagram.hpp
dr7ana: i appreciate it thank you

src/connection.cpp
dr7ana: noted and fixed

src/connection.cpp
dr7ana: ill re-add it

src/connection.cpp
dr7ana: good point thank you

src/connection.cpp
dr7ana: good catch thats an old comment

jagerman (approved):

tewinget (approved):

PR #26, opened 2023-07-27
Datagram speedtesting

dr7ana:
- implemented datagram speedtest binaries in `tests/`
- optimized send logic and dropped superfluous flags (checksum, hash, etc)
- datagrams are fast


tests/dgram-speed-client.cpp
```diff
@@ -0,0 +1,229 @@
...
+    auto client_tls = GNUTLSCreds::make(key, cert, server_cert);
+
+    send_data* d_ptr;
+
+    stream_close_callback stream_closed = [&](Stream& s, uint64_t errcode) {
+        size_t i = s.stream_id() >> 2;
+        log::critical(test_cat, "Stream {} (rawid={}) closed (error={})", i, s.stream_id(), errcode);
+    };
+
+    dgram_data_callback recv_dgram_cb = [&](bstring data) {
```
jagerman: Looking at this as a sample app implementation, it seems that we have a couple issues with the argument passed here:
- when we don't have a split packet (because either we don't have packet splitting, or we receive a small packet) we are still allocating a `bstring` that we could avoid.
- there's no way to tell *which* connection sent me this datagram.  I was looking to make this server persistent, but right now I can't which is going to be a problem when this is used in a many-to-one client-to-server architecture.

The second issue is easy to fix (just pass the connection_interface along), but the first issue is trickier because *sometimes* we want to pass a `bstring` that we already have on hand (i.e. from reassembling a split packet), but other times we we don't.

include/quic/datagram.hpp
```diff
@@ -9,9 +9,45 @@ namespace oxen::quic
...
+
+        template <typename CharType>
+        void reply(std::basic_string<CharType>&& data)
+        {
+            auto keep_alive = std::make_shared<std::basic_string<CharType>>(std::move(data));
+            std::basic_string_view<CharType> view{*keep_alive};
+            reply(view, std::move(keep_alive));
+        }
+
+        void reply(bstring_view data, std::shared_ptr<void> keep_alive = nullptr);
```
tewinget, line 46: This should probably be renamed as it is used for a non-reply send as well as a reply.

dr7ana: That's valid, but from the perspective of the callback, having "reply" somewhere makes it clear that it's directly back along that connection

From that datagram interface, the user has a getter method to retrieve the connection interface and call `send_datagram` from there

jagerman: I think `reply` is fine, but should the `send_datagram` in here taking a vector of char's also be renamed to `reply`?

dr7ana: That's a good point, I just used the same templating we had on connection_interface and stream for consistency. Maybe someone wants to send a big reply?

src/messages.cpp
```diff
@@ -175,8 +175,10 @@ namespace oxen::quic
             return {*payload, pload_id, -1, true};
         else if (addendum && not payload)
             return {*addendum, *add_id, 1, true};
+        else if (b)
+            return {*payload, pload_id, -1, false};
         else
-            return {(b ? *payload : *addendum), (b ? pload_id : *add_id), (b ? -1 : 1), false};
```
dr7ana, line 179: RIP double inline ternary

tests/dgram-speed-client.cpp
dr7ana: Resolving this per off-github discussion, to be addressed in #29 ongoing

jagerman (approved):

include/quic/connection.hpp
```diff
@@ -17,8 +17,12 @@
 #include "types.hpp"
 #include "utils.hpp"
 
+extern std::atomic<bool> datagram_test_enabled;
```
jagerman: This should be present only in a debug build

include/quic/connection.hpp
```diff
@@ -45,15 +49,19 @@ namespace oxen::quic
...
+        std::atomic<int> datagram_flip_flip_counter{0};
     };
-
 #endif
 
-    class connection_interface
+    class perf_testing
+    {
+      public:
+    };
```
jagerman: What is this?

include/quic/opt.hpp
```diff
@@ -48,6 +51,7 @@ namespace oxen::quic::opt
         Splitting mode = Splitting::NONE;
         // Note: this is the size of the entire buffer, divided amongst 4 rows
         int bufsize = 4096;
+        // If setr to
```
jagerman: ?

include/quic/utils.hpp
```diff
@@ -42,6 +42,7 @@ extern "C"
 #include <string>
 #include <string_view>
 #include <unordered_set>
+#include <variant>
```
jagerman: Not used in this header, afaict?

src/connection.cpp
```diff
@@ -32,6 +32,8 @@ extern "C"
 #include "stream.hpp"
 #include "utils.hpp"
 
+std::atomic<bool> datagram_test_enabled = false;
+
```
jagerman: NDEBUG

tests/CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ target_link_libraries(alltests PUBLIC quic Catch2::Catch2 CLI11::CLI11)
 #    test_multiclient.cpp)
 #target_link_libraries(test_multiclient PUBLIC quic)
 
+foreach(x dgram-speed-client dgram-speed-server)
+    add_executable(${x} ${x}.cpp)
+    target_link_libraries(${x} PRIVATE quic CLI11::CLI11)
+endforeach()
+
```
jagerman: Would be simpler to just add these two executables into the loop just below

include/quic/connection.hpp
dr7ana: but we dont speedtest on debug builds?

PR #28, opened 2023-07-31
Lokinet cmake/CI fix

dr7ana:
- libquic doesn't need to build oxen-encoding when it's integrated into lokinet
- 004 needs to account for hardware differences when opening streams
- 007 cannot strictly predict ngtcp2 behavior to fine granularity across different architectures

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -57,7 +57,9 @@ set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
 add_subdirectory(oxen-logging)
 
 # oxenc
-system_or_submodule(OXENC oxenc liboxenc>=1.0.4 oxen-encoding)
+if (NOT TARGET oxenc)
+    system_or_submodule(OXENC oxenc liboxenc>=1.0.4 oxen-encoding)
+endif()
 
 # libevent
 if(NOT TARGET libevent::core)
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -50,11 +50,15 @@ add_library(ngtcp2::crypto
     ngtcp2_crypto)  # ngtcp2 crypto (and implicitly base ngtcp2)
 
 # oxen-logging
-if(BUILD_STATIC_DEPS)
-    set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+if (NOT TARGET oxen::logging)
+    if(BUILD_STATIC_DEPS)
+        set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+    endif()
+    set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
+    add_subdirectory(oxen-logging)
+else()
+    set(OXEN_LOGGING_SOURCE_ROOT "${OXEN_LOGGING_SOURCE_ROOT};${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
 endif()
-set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
-add_subdirectory(oxen-logging)
 
 # oxenc
 if (NOT TARGET oxenc)
--- a/tests/004-streams.cpp
+++ b/tests/004-streams.cpp
@@ -214,11 +214,9 @@ namespace oxen::quic::test
 
         for (int i = 0; i < n_streams; ++i)
         {
-            client_endpoint->call([&, i]() {
-                streams[i] = conn_interface->get_new_stream();
-                streams[i]->send(msg);
-                send_promises[i].set_value(true);
-            });
+            streams[i] = conn_interface->get_new_stream();
+            streams[i]->send(msg);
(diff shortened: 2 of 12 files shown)


external/CMakeLists.txt
```diff
@@ -57,7 +57,9 @@ set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
 add_subdirectory(oxen-logging)
```
dr7ana: wrap this in if (NOT TARGET ...

external/CMakeLists.txt
```diff
@@ -50,14 +50,18 @@ add_library(ngtcp2::crypto
     ngtcp2_crypto)  # ngtcp2 crypto (and implicitly base ngtcp2)
 
 # oxen-logging
-if(BUILD_STATIC_DEPS)
-    set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+if (NOT TARGET oxen-logging)
```
jagerman: `oxen-logging` is a semi-internal target for oxen-logging; the public (ALIAS) target is `oxen::logging`

```suggestion
if (NOT TARGET oxen::logging)
```

external/CMakeLists.txt
```diff
@@ -50,14 +50,18 @@ add_library(ngtcp2::crypto
...
-if(BUILD_STATIC_DEPS)
-    set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+if (NOT TARGET oxen::logging)
+    if(BUILD_STATIC_DEPS)
+        set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+    endif()
+    set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
+    add_subdirectory(oxen-logging)
 endif()
-set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
```
jagerman, line 56: If it's already set up we should still append our source root to the current value (with `;` to separate)

tests/004-streams.cpp
```diff
@@ -153,24 +153,31 @@ namespace oxen::quic::test
 
         std::atomic<int> index{0};
         std::atomic<int> data_check{0};
-        opt::max_streams max_streams{8};
-        std::vector<std::shared_ptr<Stream>> streams{12};
+        int n_threads = (std::thread::hardware_concurrency() < 12) ? std::thread::hardware_concurrency() : 12;
+
+        if (n_threads < 8)
+            SKIP("Current hardware does not support ad hoc creation of > 8 threads!");
```
jagerman: This isn't necessary; you can still create however many threads you want, even if you don't have the CPUs cores/threads to back them (they just got scheduled across the cores you do have).

external/CMakeLists.txt
```diff
@@ -50,14 +50,20 @@ add_library(ngtcp2::crypto
...
-if(BUILD_STATIC_DEPS)
-    set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+if (NOT TARGET oxen::logging)
+    if(BUILD_STATIC_DEPS)
+        set(OXEN_LOGGING_FORCE_SUBMODULES ON CACHE INTERNAL "")
+    endif()
+    set(OXEN_LOGGING_SOURCE_ROOT "${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
+    add_subdirectory(oxen-logging)
+else()
+    set(OXEN_LOGGING_SOURCE_ROOT "${OXEN_LOGGING_SOURCE_ROOT};${PROJECT_SOURCE_DIR}" CACHE INTERNAL "")
```
jagerman, line 60: Need to investigate whether this appends multiple times if you re-run cmake multiple times from an existing build directory.  (The final value will end up in the build dir's CMakeCache.txt).

jagerman (approved):

tests/004-streams.cpp
```diff
@@ -209,47 +212,40 @@ namespace oxen::quic::test
...
         // 4) check the last 4
-        for (int i = 8; i < 12; ++i)
+        for (int i = n_streams - 4; i < n_streams; ++i)
             REQUIRE(receive_futures[i].get());
 
         // 5) open 2 more streams and send
         for (int i = 0; i < 2; ++i)
         {
-            std::thread open_thread([&]() {
+            client_endpoint->call([&]() {
```
jagerman: A `call` here shouldn't be needed (and, in fact, if we need it, then something is wrong in the API); we ought to be able to just directly call the code here, from the main thread.

tests/004-streams.cpp
```diff
@@ -209,47 +212,40 @@ namespace oxen::quic::test
 
         REQUIRE(tls_future.get());
 
-        // 1) open 12 streams and send
-        for (int i = 0; i < 12; ++i)
+        for (int i = 0; i < n_streams; ++i)
         {
-            std::thread stream_thread([&]() {
+            client_endpoint->call([&]() {
```
jagerman: `call` shouldn't be needed

jagerman: Neither call nor std::thread are needed here.

tests/004-streams.cpp
jagerman: Neither call nor thread are wanted here.

PR #31, opened 2023-08-01
Add `call_get` to abstract the promise-future-getter pattern

jagerman:


tewinget:
Was implicitly merged by #28

PR #32, opened 2023-08-02
[WIP] ZMQ Bridge

dr7ana:


src/CMakeLists.txt
```diff
@@ -1,3 +1,22 @@
+option(LIBQUIC_RECVMMSG "Use recvmmsg when receiving UDP packets" ${libquic_recvmmsg_default})
+option(ENABLE_PERF_TESTING "Enable optional performance testing binary logic" OFF)
+option(ENABLE_ZMQ_BRIDGE "Enable optional header-only zmq messaging bridge" OFF)
```
jagerman, line 3: In the current implementation this option isn't really header-only because of the `define` reaching into and changing `connection.hpp/cpp`.  We need to figure out a way to not touch anything at all (which might mean pushing some additional requirement, for instance of maintaining the bridge object, onto the application), so that an application can just include the header to get at the functionality if it wants it, but if it doesn't include the header then it doesn't get it.

The definition touching the `connection` class is particularly problematic because it essentially means that you have two different possible versions of the `connection` class: one with the `zmq` member and one without, and that means that you *can't* compile for the library properly unless the application carries the same compilation defintions: for example lokinet when compiling against libquic would have to know somehow that the libquic.so it is trying to link against (which might be a system-installed package) was built with or without `-DLIBQUIC_ZMQ_BRIDGE`, and will have to build with the same option.

src/connection.cpp
```diff
@@ -1129,6 +1136,10 @@ namespace oxen::quic
         datagrams = std::make_unique<DatagramIO>(*this, _endpoint, ep.dgram_recv_cb);
         pseudo_stream = std::make_shared<Stream>(*this, _endpoint, -1);
 
+#ifdef LIBQUIC_ZMQ_BRIDGE
+        zmq = _endpoint.zmq_bridge->deploy_worker(*this);
+#endif
```
jagerman: This should be opt-in at run-time.

PR #33, opened 2023-08-07
Close Handling

dr7ana:
Re-doing how `Network` is shut down both gracefully and immediately

Close handling looked like it became its own focus in #32, so pushed it to its own PR

### DIFF
--- a/include/quic/connection.hpp
+++ b/include/quic/connection.hpp
@@ -150,10 +150,12 @@ namespace oxen::quic
         bool is_inbound() const { return dir == Direction::INBOUND; }
         bool is_outbound() const { return dir == Direction::OUTBOUND; }
 
+        void halt_events();
         bool is_closing() const { return closing; }
-        void call_closing();
+        void set_closing() { closing = true; }
         bool is_draining() const { return draining; }
-        void drain() { draining = true; }
+        void set_draining() { draining = true; }
+        void call_close_cb();
 
         const ConnectionID& scid() const override { return _source_cid; }
         const ConnectionID& dcid() const { return _dest_cid; }
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -233,8 +233,7 @@ namespace oxen::quic
             log::trace(log_cat, "Note: CID-{} in closing period; signaling endpoint to delete connection", scid());
 
             _endpoint.call([this]() {
-                log::debug(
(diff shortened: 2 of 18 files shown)


tewinget (approved):

src/connection.cpp
```diff
@@ -221,6 +221,14 @@ namespace oxen::quic
         return "{:02x}"_format(fmt::join(std::begin(data), std::begin(data) + datalen, ""));
     }
 
+    void Connection::halt_events()
+    {
+        log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
+        event_del(packet_io_trigger.get());
+        event_del(packet_retransmit_timer.get());
```
jagerman: These are `event_ptr` objects which which calls event_del on destruction, so should just be `.reset()` instead of having `event_del` called directly.

include/quic/connection.hpp
```diff
@@ -150,10 +150,12 @@ namespace oxen::quic
         bool is_inbound() const { return dir == Direction::INBOUND; }
         bool is_outbound() const { return dir == Direction::OUTBOUND; }
 
+        void halt_events();
         bool is_closing() const { return closing; }
-        void call_closing();
+        void set_is_closing() { closing = true; }
```
jagerman: ~~Should this be `private:`?~~

Nevermind, this is in `connection` already which is internal.

tests/001-handshake.cpp
```diff
@@ -182,7 +175,6 @@ namespace oxen::quic::test
             auto conn_interface = client_endpoint->connect(client_remote, client_tls);
 
             REQUIRE(tls_future.valid());
-            test_net.close();
         };
```
jagerman, line 181: Maybe set the close-immediate flag on *some* of these so that it is getting used a few times in the test suite?

dr7ana: Good point, i can make a specific test case for that

include/quic/connection.hpp
jagerman: Also perhaps `set_is_closing` -> `set_closing`?

dr7ana: All those methods are called from endpoint, so we can friend class it or keep them public?

jagerman: Yeah keep them public.

PR #34, opened 2023-08-22
Do not merge - Bparser lokinet integration sandboxing

dr7ana:


include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+    {
+        return std::basic_string<std::byte>(reinterpret_cast<const std::byte*>(__str), __len);
+    }
+
+    enum class request_type
+    {
+        INCOMPLETE = 0, REQUEST = 1, RESPONSE = 2, COMMAND = 3, ERROR = 4
+    };
+
+    inline request_type translate_req_type(std::byte b)
```
jagerman: `translate` -> `decode`

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+            case std::byte{'R'}:
+                return request_type::RESPONSE;
+            case std::byte{'E'}:
+                return request_type::ERROR;
+            default:
+                return request_type::INCOMPLETE;
+        }
+    }
+
+    inline std::byte revert_req_type(request_type rt)
```
jagerman: `revert` -> `encode`

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+                throw std::invalid_argument{"This method is for construction of outgoing requests!"};
+        }
+    }
+
+    struct message
+    {
+        bstring data{};
+        bstring endpoint{};
+        bstring req_id{};
+        bstring req_body{};
```
jagerman: These don't need `{}` since std::string is a non-trivial type (that is: without the `{}` is still gets default-initialized, because it has non-trivial construction).

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+    };
+
+    struct request_t
+    {
+        // parsed request data
+        bstring data{};
+        bstring_view data_view{};
+        bstring_view req_id{};
+        bstring_view req_endpoint{};
+        bstring_view req_body{};
```
jagerman: Unnecessary `{}`

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+        size_t total_len{};
+        // the total length currently received 
+        size_t recv_len{};
+        size_t currently_read{0};
+        bool is_complete{false};   // this is a dumb name
+        request_type type{request_type::INCOMPLETE};
+
+        time_point req_time;
+        time_point timeout;
+
```
jagerman: There's a mix of initialization styles among these variables: some have no initializer, some have explicit default initializer, some have explicit default values.  The first two styles get initialized to 0, the latter doesn't.  I'd pick just one, at least for primitive variables.  (You could also use `size_t currently_read = 0;` which would work exactly the same as the `var x{};` and `var x{0};` for primitive types)

This is a personal style thing, so don't feel you need to use my preferences: I tend towards:
- `TYPE var = 123;'` for primitive types (ints, bools) and literal types (""s, ""sv, the various stl::chrono duration types), and for values where I would be calling a factory function to get the value.
- `Type var{...};' for a type where I explicitly want to invoke a constructor (which means the Type is non-primitive).
- `Type var;` for non-primitive types with a default-constructor, and for types that required construction dependent on the containing class's constructor arguments.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+        explicit request_t() = default;
+
+      private:
+        bool parse_length()
+        {
+            auto pos = data.find_first_of(std::byte{':'});
+
+            // request is incomplete with no readable request length
+            if (data.at(pos) == data.back())
+                return false;
```
jagerman: If we enter this condition, we also need to make sure that `data` only contains integers: if it contains anything else then the request is invalid and we should drop this stream with an application error code.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+
+            total_len = std::atoi(
+                    reinterpret_cast<const char*>(data_view.substr(0, pos).data()));
```
jagerman: `atoi` does "helpful" crap that we don't want to allow, like skipping whitespace, and is really meant to parse input from a dumb human.  It also *doesn't* have error conditions (i.e. if there isn't an integer there it's undefined behaviour what atoi returns—probably 0 in practice, but could be anything).

`std::from_chars` here is a much better choice for a protocol because it is more precise in what it allows and how it fails.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+
```
jagerman, line 139: If `total_len` is larger than some threshold we should abort the stream with an error.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+            // erase outer encoding
+            data.erase(0, pos+1);
```
jagerman: This is a fairly heavy operation, because all of the characters in the string starting at `pos+1` have to be moved back to the beginning of the string.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+            data_view.remove_prefix(pos);
```
jagerman: `data_view` is sort of invalidated by the `data.erase`, because the erase will shift the chars in the string, but data_view will still be pointing at the same positions in memory as it was before, which means it is still viewing from the beginning of `data`, but views beyond the end (since `data` is now shorter).

Consider:
```
    std::string data = "abcdefghijklmnopqrstuvwxyz";  
    std::string_view data_view{data};    
                                         
    data.erase(0, 3);                    
    std::cout << "data: " << data << "\n" << "view: " << data_view << "\n";  
    std::cout << "view hex: " << oxenc::to_hex(data_view) << "\n";  
```
prints:
```
data: defghijklmnopqrstuvwxyz
view: defghijklmnopqrstuvwxyz\0yz
view hex: 6465666768696a6b6c6d6e6f707172737475767778797a00797a
```
(note the null and leftover yz in the view).

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+            recv_len -= total_len + 1;  // remove one extra for the colon
+
+            data.reserve(total_len);
+            // set currently_read to 1 after req_len colon delimiter
+            currently_read = pos + 1;
+
+            return true;
+        }
+
+        bool parse_type()
```
jagerman: Discussed out of band: drop these, get a whole message, then switch to `bt_list_consumer` to parse the request once we have the full thing.

include/quic/parser.hpp
```diff
@@ -0,0 +1,454 @@
...
+                    match(current_req) :
+                    current_req->to_message(stream);
+
+            current_req.reset();
+            return msg;
+        }
+
+        std::shared_ptr<request_t> make_request(bstring endpoint, bstring body, request_type rtype)
+        {
+            // request type
```
jagerman: Easier/less maintenance to use `bt_list_producer` here.

include/quic/parser.hpp
dr7ana: good point, thank you!

jagerman:
Closing (more or less replaced by #35)

PR #39, opened 2023-09-05
init gnutls from ed keypair

tewinget:
Initialize GNUTLS creds from Ed keypair and confirm key allowed on incoming connection.

### DIFF
--- a/include/quic/gnutls_crypto.hpp
+++ b/include/quic/gnutls_crypto.hpp
@@ -9,11 +9,12 @@ extern "C"
 #include <ngtcp2/ngtcp2_crypto_gnutls.h>
 }
 
+#include <optional>
+
 #include "crypto.hpp"
 
 namespace oxen::quic
 {
-
     namespace fs = std::filesystem;
 
     using gnutls_callback = std::function<int(
@@ -22,6 +23,12 @@ namespace oxen::quic
             unsigned int when,
             unsigned int incoming,
             const gnutls_datum_t* msg)>;
+
+    constexpr auto GNUTLS_KEY_SIZE = 32;  // for now, only supporting Ed25519 keys (32 bytes)
+    using gnutls_key = std::array<unsigned char, GNUTLS_KEY_SIZE>;
+
+    // arguments: remote pubkey, is_relay
+    using gnutls_key_verify_callback = std::function<bool(const gnutls_key&, bool)>;
 
     struct gnutls_callback_wrapper
     {
@@ -69,7 +76,10 @@ namespace oxen::quic
             else
             {
                 path = NULL;
-                mem = {(uint8_t*)input.data(), (uint8_t)input.size()};
+
+                // uint32_t cast to appease narrowing conversion gods,
+                // if cert size won't fit in 32 bits we have bigger problems
+                mem = {(uint8_t*)input.c_str(), (uint32_t)input.size()};
                 format = !("-----"s.compare(input.substr(0, 5))) ? GNUTLS_X509_FMT_PEM : GNUTLS_X509_FMT_DER;
                 from_mem = true;
             }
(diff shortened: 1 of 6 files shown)


tests/008-ed-keys.cpp
```diff
@@ -0,0 +1,147 @@
...
+        SECTION("Successful Connection")
+        {
+            server_tls->set_key_verify_callback([&](const auto& key, bool is_relay) {
+                key_was_allowed.set_value(true);
+                return true;
+            });
+
+            REQUIRE_NOTHROW(client_endpoint->connect(client_remote, client_tls));
+
+            REQUIRE(f.wait_for(20ms) == std::future_status::ready);
```
dr7ana: Are these sleeps being used to give the client time to establish a connection? You could use the same logic in the other tests like [001](https://github.com/oxen-io/oxen-libquic/blob/857188c5d21ea01819c0d88258b84f592ed05125/tests/001-handshake.cpp#L190), where the `client_tls_policy` is given a callback that signals "connection ready"

tests/008-ed-keys.cpp
```diff
@@ -0,0 +1,147 @@
...
+                    "Key mismatch:\n{}\n{}",
+                    oxenc::to_hex(key.begin(), key.end()),
+                    oxenc::to_hex(client_pubkey_array.begin(), client_pubkey_array.end()));
+
+            key_was_allowed.set_value(false);
+            return false;
+        };
+
+        client_tls->set_key_verify_callback(client_key_allowed_cb);
+        server_tls2->set_key_verify_callback([&](auto, auto) {
```
dr7ana: Since you re-use this lambda 3 times, maybe you could define it where you defined `client_key_allowed_cb` and pass the lambda to `set_key_verify_callback`?

tests/008-ed-keys.cpp
tewinget: This is a future wait with a timeout, rather than a sleep.  It's so the test fails rather than hanging indefinitely if something goes horribly wrong.

tests/008-ed-keys.cpp
tewinget: true, will do.

tewinget: note that this only removes 1 duplicate though, as the other is returning false rather than true (and not based on input to the lambda)

tests/008-ed-keys.cpp
dr7ana: We don't do this in any other test, and any failure case where it hangs indefinitely should be handled internally?

jagerman: > Are these sleeps

Not a sleep -- this just puts an upper limit on how long it can wait before returning with a non-ready status.

> We don't do this in any other test, and any failure case where it hangs indefinitely should be handled internally

We probably should -- if a failure happens in a test now the test suite hangs indefinitely, rather than failing, where failing would be nicer.

Perhaps putting this in a common test suite function that we can use in many places, something like:
```c++
template <typename F>
void require_future(F& f, std::chrono::milliseconds timeout = 1s) {
    REQUIRE(f.wait_for(timeout) == std::future_status::ready);
}
```
That said, 20ms seems way too short for this (because this is just a fail-safe in case it hangs).

include/quic/gnutls_crypto.hpp
```diff
@@ -23,6 +25,12 @@ namespace oxen::quic
             unsigned int incoming,
             const gnutls_datum_t* msg)>;
 
+    constexpr auto GNUTLS_KEY_SIZE = 32;  // for now, only supporting Ed25519 keys (32 bytes)
```
jagerman: This might be better as an explicit `size_t` rather than auto (which makes it an `int`).

include/quic/gnutls_crypto.hpp
```diff
@@ -69,7 +77,10 @@ namespace oxen::quic
             else
             {
                 path = NULL;
-                mem = {(uint8_t*)input.data(), (uint8_t)input.size()};
+
+                // uint32_t cast to appease narrowing conversion gods,
+                // if cert size won't fit in 32 bits we have bigger problems
+                mem = {(uint8_t*)input.c_str(), (uint32_t)input.size()};
```
jagerman:
```suggestion
                mem = {reinterpret_cast<uint8_t*>(input.c_str()), static_cast<uint32_t>(input.size())};
```

tests/008-ed-keys.cpp
dr7ana: > This is a future wait with a timeout, rather than a sleep

> Not a sleep 

I was using "sleep" in the notional rather than literal sense

My main issue with this is using certificate verification as a proxy for "connection ready." We know already that this is not how ngtcp2 signals that at all

tests/008-ed-keys.cpp
```diff
@@ -0,0 +1,148 @@
...
+            // REQUIRE_THROWS(GNUTLSCreds::make_from_ed_keys(client_pubkey, client_seed)); // wrong order
+        };
+
+        SECTION("Keys Load Correctly")
+        {
+            REQUIRE_NOTHROW(GNUTLSCreds::make_from_ed_keys(client_seed, client_pubkey));
+        };
+    };
+
+    TEST_CASE("008 - Ed Keys: Connections", "[001][keys][tls][connection]")
```
jagerman: Perhaps name this "Handshake" rather than "Connections"?

tewinget: agreed.

include/quic/gnutls_crypto.hpp
tewinget: so gnutls_datum_t is annoying and either requires a copy here into an allocated c-string or something *or* requires casting away const.  For simplicity, as it appears safe, this casts away const.  Could do that in the C++ way though,  I suppose.

include/quic/gnutls_crypto.hpp
tewinget: makes sense

dr7ana (approved):

PR #43, opened 2023-09-14
extra closure methods

dr7ana:
Close methods needed for lokinet integration

### DIFF
--- a/include/quic/endpoint.hpp
+++ b/include/quic/endpoint.hpp
@@ -171,6 +171,8 @@ namespace oxen::quic
 
         void close_connection(Connection& conn, int code = NGTCP2_NO_ERROR, std::string_view msg = "NO_ERROR"sv);
 
+        void close_connection(ConnectionID cid, int code = NGTCP2_NO_ERROR, std::string_view msg = "NO_ERROR"sv);
+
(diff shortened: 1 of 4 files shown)


include/quic/endpoint.hpp
```diff
@@ -171,6 +171,8 @@ namespace oxen::quic
 
         void close_connection(Connection& conn, int code = NGTCP2_NO_ERROR, std::string_view msg = "NO_ERROR"sv);
 
+        void close_connection(ConnectionID cid, int code = NGTCP2_NO_ERROR, std::string_view msg = "NO_ERROR"sv);
```
tewinget, line 174: It looks like `Connection::close_connection` is calling the previous declaration of this, so this appears unused?

dr7ana: It’s not unused, closing connections via SCID is useful for lokinet integration

tewinget: Ah okay, it's just unused in this commit; that makes more sense.

dr7ana: Oh i understand what you meant now! Yes exactly it’s for external use. Will PM you

tewinget (approved):

PR #44, opened 2023-09-19
Added Address neq operator

dr7ana:


include/quic/address.hpp
```diff
@@ -5,6 +5,8 @@
 
 namespace oxen::quic
 {
+    constexpr std::array<uint8_t, 16> _ipv6_any_addr = {0};
```
jagerman: This should be `inline constexpr`

jagerman: (Or alternatively, moved into the .cpp along with the `is_any_addr` body)

dr7ana: fixed thank you

jagerman (approved):

PR #45, opened 2023-09-19
Compile-time check for listen() having a tls credentials argument

jagerman:
This was throwing, but it's nicer to detect that failure case at compile time.

### DIFF
--- a/tests/001-handshake.cpp
+++ b/tests/001-handshake.cpp
@@ -50,7 +50,7 @@ namespace oxen::quic::test
             auto ep_notls = test_net.endpoint(default_addr);
             auto ep_tls = test_net.endpoint(local_addr);
 
-            REQUIRE_THROWS(ep_notls->listen());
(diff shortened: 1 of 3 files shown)


tests/001-handshake.cpp
```diff
@@ -50,7 +50,7 @@ namespace oxen::quic::test
             auto ep_notls = test_net.endpoint(default_addr);
             auto ep_tls = test_net.endpoint(local_addr);
 
-            REQUIRE_THROWS(ep_notls->listen());
+            // REQUIRE_THROWS(ep_notls->listen());  // Shouldn't compile if uncommented!
```
dr7ana: This change is definitely an improvement. Small detail if you like: dropping `REQUIRE_THROWS` and just leaving `ep_notls->listen()` commented out would be marginally clearer

tewinget: agreed; that hadn't even occurred to me.

tewinget (approved):

PR #47, opened 2023-09-20
Bparser integration + Stream construction/queueing

dr7ana:
Bparser is now implemented via the developed custom stream creation methods. New stream creation methods are implemented to queue and construct user-defined extensions of streams

### DIFF
--- a/include/quic/parser.hpp
+++ b/include/quic/parser.hpp
@@ -13,20 +13,10 @@ namespace oxen::quic
     const std::chrono::seconds TIMEOUT{10};
 
     // request sizes
-    inline constexpr long long MAX_REQ_LEN{10000000};
+    inline constexpr long long MAX_REQ_LEN = 10_M;
 
     // Application error
     inline constexpr uint64_t BPARSER_EXCEPTION = (1ULL << 60) + 69;
-
-    inline std::basic_string<std::byte> operator""_bs(const char* __str, size_t __len) noexcept
-    {
-        return std::basic_string<std::byte>(reinterpret_cast<const std::byte*>(__str), __len);
-    }
-
-    inline std::string_view to_sv(bstring_view x)
-    {
-        return {reinterpret_cast<const char*>(x.data()), x.size()};
-    }
 
     struct message
     {
@@ -42,20 +32,13 @@ namespace oxen::quic
         {
             oxenc::bt_list_consumer btlc(data);
 
-            try
-            {
-                req_type = btlc.consume_string_view();
-                req_id = btlc.consume_string_view();
+            req_type = btlc.consume_string_view();
+            req_id = btlc.consume_string_view();
 
-                if (req_type == "Q" || req_type == "C")
-                    endpoint = btlc.consume_string_view();
+            if (req_type == "Q" || req_type == "C")
+                endpoint = btlc.consume_string_view();
 
-                req_body = btlc.consume_string_view();
-            }
-            catch (...)
-            {
-                log::critical(bp_cat, "Invalid request body!");
-            }
+            req_body = btlc.consume_string_view();
         }
 
         std::string rid() { return std::string{req_id}; }
@@ -79,9 +62,9 @@ namespace oxen::quic
         explicit sent_request(std::string d, std::string rid) : req_id{std::move(rid)}
         {
             total_len = d.length();
+            data.reserve(data.length() + total_len);
             data += std::to_string(total_len);
             data += ':';
-            data.reserve(data.length() + total_len);
             data.append(d);
 
             req_time = get_time();
@@ -104,6 +87,8 @@ namespace oxen::quic
         std::string size_buf;
 
         size_t current_len{0};
+
+        std::atomic<int64_t> next_rid{0};
 
         std::function<void(Stream&, message)> recv_callback;
 
@@ -151,7 +136,15 @@ namespace oxen::quic
         {
             log::info(bp_cat, "bparser recv data callback called!");
             log::debug(bp_cat, "Received data: {}", buffer_printer{data});
-            process_incoming(to_sv(data));
+
+            try
+            {
+                process_incoming(to_sv(data));
+            }
+            catch (std::exception& e)
+            {
+                log::error(bp_cat, "Exception caught: {}", e.what());
+            }
         }
 
         void closed(uint64_t app_code) override
@@ -288,7 +281,7 @@ namespace oxen::quic
         std::shared_ptr<sent_request> make_request(std::string endpoint, std::string body)
         {
             oxenc::bt_list_producer btlp;
-            std::string rid = "111112222233333"s;  // replace with libsodium function call
+            std::string rid = std::to_string(++next_rid);
 
             try
             {
@@ -297,7 +290,7 @@ namespace oxen::quic
                 btlp.append(endpoint);
                 btlp.append(body);
 
-                auto req = std::make_shared<sent_request>(std::move(btlp).str(), std::move(rid));
+                auto req = std::make_shared<sent_request>(std::move(btlp).str(), rid);
                 return req;
             }
             catch (...)
@@ -311,7 +304,7 @@ namespace oxen::quic
         std::shared_ptr<sent_request> make_command(std::string endpoint, std::string body)
         {
             oxenc::bt_list_producer btlp;
-            std::string rid = "111112222233333"s;  // replace with libsodium function call
+            std::string rid = std::to_string(++next_rid);
 
             try
             {
@@ -320,7 +313,7 @@ namespace oxen::quic
                 btlp.append(endpoint);
                 btlp.append(body);
 
-                auto req = std::make_shared<sent_request>(std::move(btlp).str(), std::move(rid));
+                auto req = std::make_shared<sent_request>(std::move(btlp).str(), rid);
                 return req;
             }
             catch (...)
@@ -341,7 +334,7 @@ namespace oxen::quic
                 btlp.append(rid);
                 btlp.append(body);
 
-                auto req = std::make_shared<sent_request>(std::move(btlp).str(), std::move(rid));
+                auto req = std::make_shared<sent_request>(std::move(btlp).str(), rid);
                 return req;
             }
             catch (...)
--- a/include/quic/parser.hpp
+++ b/include/quic/parser.hpp
@@ -36,7 +36,9 @@ namespace oxen::quic
         std::string_view endpoint;
         std::string_view req_body;
 
-        message(std::string req) : data{std::move(req)}
+        bool error{false};
+
+        message(std::string req, bool is_error = false) : data{std::move(req)}, error{is_error}
         {
             oxenc::bt_list_consumer btlc(data);
 
@@ -56,7 +58,7 @@ namespace oxen::quic
             }
         }
 
-        std::string rid() { return std::move(std::string{req_id}); }
+        std::string rid() { return std::string{req_id}; }
         std::string_view view() { return {data}; }
     };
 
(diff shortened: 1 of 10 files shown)


include/quic/parser.hpp
```diff
@@ -0,0 +1,384 @@
...
+                req_id = btlc.consume_string_view();
+
+                if (req_type == "Q" || req_type == "C")
+                    endpoint = btlc.consume_string_view();
+
+                req_body = btlc.consume_string_view();
+            }
+            catch (...)
+            {
+                log::critical(bp_cat, "Invalid request body!");
```
tewinget: This should probably re-`throw` so the calling context knows something went wrong.

include/quic/parser.hpp
```diff
@@ -0,0 +1,384 @@
...
+{
+    inline auto bp_cat = oxen::log::Cat("bparser");
+
+    using time_point = std::chrono::steady_clock::time_point;
+
+    // timeout is used for sent requests awaiting responses
+    const std::chrono::seconds TIMEOUT{10};
+
+    // request sizes
+    inline constexpr long long MAX_REQ_LEN{10000000};
```
jagerman:
```suggestion
    inline constexpr long long MAX_REQ_LEN = 10_M;
```

include/quic/parser.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+    // Application error
+    inline constexpr uint64_t BPARSER_EXCEPTION = (1ULL << 60) + 69;
+
+    inline std::basic_string<std::byte> operator""_bs(const char* __str, size_t __len) noexcept
```
jagerman: Leading-double-underscore names are reserved for internal use by the c/c++ compiler.  (Also, leading-single-underscore-followed-by-a-capital-letter).

include/quic/parser.hpp
jagerman: Perhaps better: *don't* catch these, and do the try/catch further up the chain (i.e. where the message get constructed).

include/quic/parser.hpp
```diff
@@ -0,0 +1,384 @@
...
+
+                req_body = btlc.consume_string_view();
+            }
+            catch (...)
+            {
+                log::critical(bp_cat, "Invalid request body!");
+            }
+        }
+
+        std::string rid() { return std::move(std::string{req_id}); }
```
jagerman: No need for the `move` here just `return std::string{req_id};` will do the trick.

include/quic/parser.hpp
```diff
@@ -0,0 +1,378 @@
...
+
+                buf += req;
+                return;
+            }
+        }
+
+        std::shared_ptr<sent_request> make_request(std::string endpoint, std::string body)
+        {
+            oxenc::bt_list_producer btlp;
+            std::string rid = "111112222233333"s;  // replace with libsodium function call
```
jagerman: Needs fix

jagerman: `std::atomic<int64_t>` in the bparser, and use ++ on that to generate request ids

include/quic/parser.hpp
```diff
@@ -0,0 +1,378 @@
...
+        std::chrono::steady_clock::time_point timeout;
+
+        bool is_empty() const { return data.empty() && total_len == 0; }
+
+        explicit sent_request(std::string d, std::string rid) : req_id{std::move(rid)}
+        {
+            total_len = d.length();
+            data += std::to_string(total_len);
+            data += ':';
+            data.reserve(data.length() + total_len);
```
jagerman: This reserve should come before we start appending to `data` to be effective.

include/quic/parser.hpp
```diff
@@ -0,0 +1,378 @@
...
+            data.append(d);
+
+            req_time = get_time();
+            timeout = req_time + TIMEOUT;
+        }
+
+        message to_message() { return {data}; }
+
+        std::string_view view() { return {data}; }
+        std::string&& payload() { return std::move(data); }
```
jagerman:
```suggestion
        std::string&& payload() && { return std::move(data); }
```
so that you have to explicitly use `std::move(req).payload()` but just `req.payload()` wouldn't be allowed to destroy the data.

include/quic/parser.hpp
```diff
@@ -0,0 +1,378 @@
...
+        {
+            log::debug(bp_cat, "Bparser set user-provided close callback!");
+            close_callback = std::move(close_cb);
+        }
+
+        bool match(std::string_view rid)
+        {
+            log::trace(bp_cat, "{} called", __PRETTY_FUNCTION__);
+
+            for (auto& r : sent_reqs)
```
jagerman: This linear scan is a little unfortunate (e.g. think SPNS where we have thousands of requests out and arriving at once).

Thinking about what we could do instead...

dr7ana: Timestamps and request id's are both monotonically increasing, so we could binary search for log(n) complexity

tewinget: Currently is an ordered map of `request time -> request` which is good for clearing timed-out requests, but means matching request id is slower.  Could instead map `request id -> request`.  If `request id` is monotonically increasing, time should as well, so the same clear timeout logic should work (except where request id would wrap, but...64-bit integer.)

jagerman: Yeah.  Thinking about that it makes sense: we probably will have to do something along the lines of (pseudocode):

while first entry is expired:
    trigger failure callback for first entry
    pop first entry

and so for expiring, all we really care about is that they are ordered, but for lookup we're operating on the id, so id as key seems to make sense to me.

Or hell, maybe we could just go back to std::deque and do a std::binary_search that compares the `id` on lookup, and just pop off the front (as pseudocode above) when we look for expired stuff.

include/quic/parser.hpp
```diff
@@ -0,0 +1,380 @@
...
+
+                buf += req;
+                return;
+            }
+        }
+
+        std::shared_ptr<sent_request> make_request(std::string endpoint, std::string body)
+        {
+            oxenc::bt_list_producer btlp;
+            std::string rid = std::to_string(++next_rid);
```
jagerman: Probably easier just to put an int in the field, i.e.:
```suggestion
            auto rid = ++next_rid;
```
here and when we parse it, `consume_integer<int64_t>()` instead of consuming/parsing a string (in the message constructor, around line 36).

dr7ana: Nice i didn't realize there were templated int parsing/consuming arguments in oxenc, thank you

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+#include "utils.hpp"
+
+namespace oxen::quic
+{
+    inline auto bp_cat = oxen::log::Cat("bparser");
+
+    using time_point = std::chrono::steady_clock::time_point;
+
+    // timeout is used for sent requests awaiting responses
+    const std::chrono::seconds TIMEOUT{10};
```
jagerman: Can be `constexpr`

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+        //      if (not m.timed_out)
+        //      { // success logic }
+        //      ... // is identical to:
+        //      if (m)
+        //      { // success logic }
+        //  }
+        operator bool() const { return not timed_out; }
+
+        int64_t rid() { return req_id; }
+        std::string_view view() { return {data}; }
```
jagerman: These two can be `const` methods

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+            data.reserve(data.length() + total_len);
+            data += std::to_string(total_len);
+            data += ':';
+            data.append(d);
+
+            req_time = get_time();
+            timeout = req_time + TIMEOUT;
+        }
+
+        bool is_expired(std::chrono::steady_clock::time_point tp) { return timeout < tp; }
```
jagerman: Can be `const` as well

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+        {
+            log::debug(bp_cat, "{} called", __PRETTY_FUNCTION__);
+
+            auto req = make_response(rid, std::move(body), error);
+            send(req->payload());
+        }
+
+        void check_timeouts()
+        {
+            const auto& now = get_time();
```
jagerman:
```suggestion
            const auto now = get_time();
```
I think this was a reference to a temporary

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+        void receive(bstring_view data) override
+        {
+            log::info(bp_cat, "bparser recv data callback called!");
+            log::debug(bp_cat, "Received data: {}", buffer_printer{data});
+
+            try
+            {
+                process_incoming(to_sv(data));
+            }
+            catch (std::exception& e)
```
jagerman:
```suggestion
            catch (const std::exception& e)
```

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+            {
+                log::error(bp_cat, "Exception caught: {}", e.what());
```
jagerman, line 184: We need to close the stream here, because if we hit an exception then we're in an unrecoverable state on the stream and we really can't meaningfully keep parsing.

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+                    return;
+
+            } while (not sent_reqs.empty());
+        }
+
+        void receive(bstring_view data) override
+        {
+            log::info(bp_cat, "bparser recv data callback called!");
+            log::debug(bp_cat, "Received data: {}", buffer_printer{data});
+
```
jagerman, line 177: I think we need a:
```C++
if (is_closing())
    return;
```
here so that, if we closed the stream, we don't try processing more.

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+        std::chrono::steady_clock::time_point req_time;
+        std::chrono::steady_clock::time_point timeout;
+
+        bool is_empty() const { return data.empty() && total_len == 0; }
+
+        explicit sent_request(bparser& bp, std::string d, int64_t rid) : req_id{rid}, return_sender{bp}
+        {
+            total_len = d.length();
+            data.reserve(data.length() + total_len);
```
jagerman: I think just drop this reserve; it isn't quite right, and the "right" value here is annoyingly difficult to calculate (because it involves stringification of an integer).  But small-string-optimization will most likely save us anyway since the initial `"1234:"` will get written before it allocates anything.

jagerman: (I think the "correct" calculation is `1 + ceil(log10(total_len)) + d.size()`, but to hell with calculating that).

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+
+        // total length of the request; is at the beginning of the request
+        size_t total_len;
+
+        std::chrono::steady_clock::time_point req_time;
+        std::chrono::steady_clock::time_point timeout;
+
+        bool is_empty() const { return data.empty() && total_len == 0; }
+
+        explicit sent_request(bparser& bp, std::string d, int64_t rid) : req_id{rid}, return_sender{bp}
```
jagerman: Since we aren't going to keep or move the `d` variable itself, this can just be more generic `std::string_view d` parameter.

include/quic/parser.hpp
```diff
@@ -0,0 +1,421 @@
...
+                }
+                else
+                    return;
+
+            } while (not sent_reqs.empty());
+        }
+
+        void receive(bstring_view data) override
+        {
+            log::info(bp_cat, "bparser recv data callback called!");
```
jagerman: I think this is left-over debugging to remove

jagerman (approved):

tests/006-server-send.cpp
```diff
@@ -176,5 +176,16 @@ namespace oxen::quic::test
...
+        log::critical(
+                log_cat,
+                "Server: Local created ID = {}, Remote created ID = {}",
+                server_stream->stream_id(),
+                server_extracted_stream->stream_id());
+        log::critical(
+                log_cat,
+                "Client: Local created ID = {}, Remote created ID = {}",
+                client_stream->stream_id(),
+                client_extracted_stream->stream_id());
```
jagerman: Are these supposed to still be here?

PR #48, opened 2023-09-21
Fix promise sync in bool_waiter; rename to callback_waiter

jagerman:
We were setting the promise in bool_waiter *before* we call the wrapped lambda, but that introduces a race condition because we have test code that waits on the promise and then checks a value that it expects to have been set inside the lambda.

This reverses it so that the promise is set *after* the lambda is called.

While here, I also refactored this to a `promise<void>` since we are never actually meaningfully using the `bool` value (it is only ever set to `true`) and removed the `get()` function (since the only difference between it and `wait_ready()` is that `get()` can hang forever), and then since it isn't a "bool" waiter anymore, renamed the whole thing to callback_waiter.

Edit: also renamed `wait_ready` to `wait`

Also documented its usage.

### DIFF
--- a/tests/utils.hpp
+++ b/tests/utils.hpp
@@ -119,37 +119,69 @@ namespace oxen::quic
     template <typename Class, typename Ret, typename... Args>
     struct functional_helper<Ret (Class::*)(Args...) const>
     {
+        using return_type = Ret;
+        static constexpr bool is_void = std::is_void_v<Ret>;
         using type = std::function<Ret(Args...)>;
     };
 
     template <typename T>
     using functional_helper_t = typename functional_helper<T>::type;
 
+    struct set_on_exit
+    {
+        std::promise<void>& p;
+        explicit set_on_exit(std::promise<void>& p) : p{p} {}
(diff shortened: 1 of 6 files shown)


tests/utils.hpp
```diff
@@ -119,37 +119,69 @@ namespace oxen::quic
...
+    /// Usage example:
+    ///
+    ///     int foo = 0;
+    ///     callback_waiter waiter{[&foo](int a, int b) { foo = a + b; }};
+    ///     invoke_something(waiter);
+    ///
+    /// where `invoke_something` takes a `std::function<int(Foo&, int)>`.  The test code would then
+    /// go on to synchronize with:
+    ///
+    ///     REQUIRE(waiter.wait_ready()); // will fail if the lambda doesn't get called with ~1s
```
tewinget:
```suggestion
    ///     REQUIRE(waiter.wait_ready(/* 1s */)); // will fail if the lambda doesn't get called within ~1s
```

tewinget (approved):

jagerman:
Rebased on top of #47.

PR #50, opened 2023-09-25
Move ALPN code to Endpoint where it belongs

tewinget:
Also adds tests for ALPNs

Adds configurable connection handshake timeout

Reworks tests/008-ed-keys to no longer care about ALPNs

### DIFF
--- a/include/quic/endpoint.hpp
+++ b/include/quic/endpoint.hpp
@@ -41,6 +41,9 @@ namespace oxen::quic
     {
       private:
         void handle_ep_opt(opt::enable_datagrams dc);
+        void handle_ep_opt(opt::outbound_alpns alpns);
+        void handle_ep_opt(opt::inbound_alpns alpns);
+        void handle_ep_opt(opt::handshake_timeout timeout);
         void handle_ep_opt(dgram_data_callback dgram_cb);
         void handle_ep_opt(connection_open_callback conn_established_cb);
         void handle_ep_opt(connection_closed_callback conn_closed_cb);
@@ -123,7 +126,13 @@ namespace oxen::quic
                         if (auto [itr, success] = conns.emplace(ConnectionID::random(), nullptr); success)
                         {
                             itr->second = Connection::make_conn(
-                                    *this, itr->first, ConnectionID::random(), std::move(path), outbound_ctx);
+                                    *this,
+                                    itr->first,
+                                    ConnectionID::random(),
+                                    std::move(path),
+                                    outbound_ctx,
+                                    outbound_alpns,
+                                    handshake_timeout);
 
                             p.set_value(itr->second);
                             return;
@@ -226,6 +235,10 @@ namespace oxen::quic
 
         std::shared_ptr<IOContext> outbound_ctx;
         std::shared_ptr<IOContext> inbound_ctx;
+
+        std::vector<std::string> outbound_alpns;
+        std::vector<std::string> inbound_alpns;
+        std::chrono::nanoseconds handshake_timeout{5s};
 
         void _init_internals();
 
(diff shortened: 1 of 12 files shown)


include/quic/endpoint.hpp
```diff
@@ -41,6 +41,9 @@ namespace oxen::quic
     {
       private:
         void handle_ep_opt(opt::enable_datagrams dc);
+        void handle_ep_opt(opt::outbound_alpns alpns);
+        void handle_ep_opt(opt::inbound_alpns alpns);
+        void handle_ep_opt(opt::handshake_timeout timeout);
```
dr7ana, line 46: I feel like this might make more sense in `IOContext`, where it would still be propogated across all connections maintained by the Endpoint

src/gnutls_crypto.cpp
```diff
@@ -266,35 +267,39 @@ namespace oxen::quic
...
             }
-
-            if (auto rv = gnutls_alpn_set_protocols(session, proto, alpn_count, 0); rv < 0)
+        }
+        else  // set default, mandatory ALPN string
+        {
+            gnutls_datum_t alpn;
+            auto default_alpn = "default"s;
+            alpn.data = reinterpret_cast<uint8_t*>(default_alpn.data());
+            alpn.size = default_alpn.size();
```
dr7ana: Minor (minor) optimization: You could have this hard-coded as a const member variable or something similar?

src/gnutls_crypto.cpp
```diff
@@ -342,6 +347,18 @@ namespace oxen::quic
...
+    std::string_view GNUTLSSession::selected_alpn()
+    {
+        gnutls_datum_t proto;
+        if (auto rv = gnutls_alpn_get_selected_protocol(session, &proto); rv < 0)
+        {
+            log::warning(log_cat, "{} called, but ALPN negotiation incomplete.", __PRETTY_FUNCTION__);
+            return ""sv;
+        }
+
+        return proto.size ? std::string_view{(const char*)proto.data, proto.size} : ""sv;
```
dr7ana, line 355: Maybe return an optional instead? Might make the return logic and return verification in the calling scope simpler

This is making me wonder if it's worth putting in the time to wrap `gnutls_datam_t` in a lightweight wrapper similar to the artist formerly known as `datum` (currently `x509_loader`)

tests/010-alpns.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+        // this has to destroy *after* network, in case it doesn't go off before then
+        auto client_closed = callback_waiter{[](connection_interface&, uint64_t) {}};
+
+        Network test_net{};
+
+        SECTION("No ALPNs specified")
+        {
+            auto server_endpoint = test_net.endpoint(server_local, timeout);
+            REQUIRE(server_endpoint->listen(server_tls));
```
dr7ana, line 47: Thought: I think there's a strong argument to make a lot of our REQUIRE's into CHECK's. It makes sense to me to save REQUIRE for critical targeted functionalities of the unit tests themselves

dr7ana (approved):
Looks good to me! Left a few comments for things we can discuss here or off band

src/gnutls_crypto.cpp
tewinget: yeah, fair point.  Will do.

src/gnutls_crypto.cpp
tewinget: Could use `optional` here, but I tend to prefer to use a default "empty" state if the return type has that concept, which in this case an empty string represents just fine I think.  If I were to switch it to `optional`, I would probably change the return inside the `if` to return `nullopt` and leave this return as the empty string, as the former implies it is not yet chosen...but actually I should probably make it throw a `logic_error` in that case because this shouldn't be called before the ALPN negotiation is complete anyway.

tests/010-alpns.cpp
tewinget: Agreed, there are many cases where `CHECK` would be better.  This one specifically is required though, as the section cannot continue if this fails.

PR #51, opened 2023-09-26
Revisit bt request stream api

dr7ana:
- commands are now registered and methods are executed automatically
- now only commands; providing a callback makes it implicitly a request

### DIFF
--- a/src/btstream.cpp
+++ b/src/btstream.cpp
@@ -1,5 +1,7 @@
 #include "btstream.hpp"
 
+#include "endpoint.hpp"
+
 namespace oxen::quic
 {
     message::message(BTRequestStream& bp, std::string req, bool is_error) :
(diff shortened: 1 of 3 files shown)


src/btstream.cpp
```diff
@@ -108,25 +108,9 @@ namespace oxen::quic
...
-
-        if (itr != sent_reqs.end() and itr->get()->req_id == rid)
-        {
-            log::debug(bp_cat, "Successfully matched response to sent request!");
-            sent_reqs.erase(itr);
-            return true;
-        }
-
-        return false;
+        func_map.emplace(std::move(endpoint), std::move(func));
```
jagerman: Maybe throw if this gives back `inserted == false`?

jagerman: Or:
```c++
func_map[std::move(endpoint)] = std::move(func);
```
to replace it.

jagerman (approved):

PR #53, opened 2023-09-28
configurable request timeout

dr7ana:


include/quic/btstream.hpp
```diff
@@ -10,7 +10,7 @@ namespace oxen::quic
     using time_point = std::chrono::steady_clock::time_point;
 
     // timeout is used for sent requests awaiting responses
-    inline constexpr std::chrono::seconds TIMEOUT{10};
+    inline constexpr std::chrono::milliseconds TIMEOUT{10s};
```
tewinget: I think a lower default of like 2s would be better here, and if a message wants a longer timeout it can specify.

include/quic/btstream.hpp
```diff
@@ -67,26 +67,39 @@ namespace oxen::quic
...
+        template <typename... Opt>
+        sent_request(BTRequestStream& bp, std::string_view d, int64_t rid, Opt&&... opts) :
+                req_id{rid}, return_sender{bp}, req_time{get_time()}, expiry{req_time}
+        {
+            ((void)handle_req_opts(std::forward<Opt>(opts)), ...);
 
-        bool is_expired(std::chrono::steady_clock::time_point tp) const { return timeout < tp; }
+            total_len = d.length();
+            data = oxenc::bt_serialize(d);
+            expiry += (timeout) ? *timeout : TIMEOUT;
```
jagerman:
```suggestion
            expiry += timeout.value_or(TIMEOUT);
```

include/quic/btstream.hpp
tewinget: also, since this is constexpr, it should probably be named something like `DEFAULT_TIMEOUT`

jagerman: No need to change this; a `seconds` is implicitly convertible to a `milliseconds` (or other more-precise types).

jagerman: Also perhaps:
```suggestion
    inline constexpr auto TIMEOUT = 10s;
```

include/quic/btstream.hpp
```diff
@@ -143,8 +187,27 @@ namespace oxen::quic
...
+                btlp.append(rid);
+                btlp.append(endpoint);
+                btlp.append(body);
+
+                return std::make_shared<sent_request>(*this, std::move(btlp).str(), rid, std::forward<Opt>(opts)...);
+            }
+            catch (const std::exception& e)
+            {
+                log::critical(bp_cat, "Invalid outgoing command encoding: {}", e.what());
+                return nullptr;
```
jagerman:
```suggestion
                throw;
```
so that we log, but we still propagate the original exception.

include/quic/btstream.hpp
```diff
@@ -120,7 +134,34 @@ namespace oxen::quic
...
+        void command(std::string endpoint, std::string body, Opt&&... opts)
+        {
+            log::trace(bp_cat, "{} called", __PRETTY_FUNCTION__);
+
+            auto req = make_command(std::move(endpoint), std::move(body), std::forward<Opt>(opts)...);
+
+            if (req->cb)
+            {
+                send(req->view());
+                sent_reqs.push_back(std::move(req));
```
jagerman, line 160: I think we might need to loop-call this to make the `sent_reqs` access thread safe?  (And while we're there, probably worthwhile swapping the order of the two statements)

jagerman: send(sent_reqs.emplace_back(std::move(req))->view());

jagerman (approved):

PR #56, opened 2023-10-19
Add non-public range lists

jagerman:
This allows us to do a check for whether an address is public from new Lokinet code that is transitioning to use oxen::quic::Address in some places rather than Lokinet's painful `hnuintnuhuh_t` types.

See llarp/net/bogon_ranges.hpp for the current lokinet equivalent.  This also fixes some omissions in the IPv6 ranges (e.g. 0xfc00/8 in lokinet is incorrect and should be a /7).

### DIFF
--- a/include/quic/address.hpp
+++ b/include/quic/address.hpp
@@ -62,7 +62,30 @@ namespace oxen::quic
             return *this;
         }
 
+        // Converts an IPv4 address into an IPv4-mapped IPv6 address.  The address must already be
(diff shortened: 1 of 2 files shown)


tewinget (approved):

tewinget:
Forgot to look at CI before clicking approve...well the change set *looks* good, pending CI passing.

dr7ana (approved):

PR #60, opened 2023-11-02
Add packet batch trigger for better packet handling

jagerman:
This adds an optional packet batch callback to endpoint that allows endpoint to signal the application when it is done processing a set of incoming packets for the endpoint -- either because it has processed them all, or because it hit the max-recv-per-loop limit (64).

This is designed to allow an application to more efficiently deal with packets, especially datagrams, in batches: by using this an application can use the datagram handler to collect incoming datagrams, then use the batch callback to process accumulated datagrams in one go.  Because the batch callback always fires *before* libquic goes back to potentially blocking waiting to read new packets, this means the application can do batch processing without needing to resort to timers or polling for packet handling.

This is particularly important for Lokinet, where we take the packet in the callback transfer it to a job in the Lokinet main loop thread to process it there: doing this one packet at a time is not likely to scale well because of the sheer number of jobs that would have to be put on the logic thread event queue; by batching them into groups of up to 64 packets at a time we ought to be able to do considerably better.

### DIFF
--- a/include/quic/udp.hpp
+++ b/include/quic/udp.hpp
@@ -38,6 +38,7 @@ namespace oxen::quic
                 ;
 
         using receive_callback_t = std::function<void(const Packet& pkt)>;
+        using post_receive_callback_t = std::function<void()>;
 
         UDPSocket() = delete;
 
@@ -45,10 +46,23 @@ namespace oxen::quic
         /// binding to an any address (or any port) you can retrieve the realized address via
         /// address() after construction.
         ///
-        /// When packets are received they will be fed into the given callback.
+        /// When packets are received they will be fed into the given `on_receive` callback.
+        ///
+        /// The optional `post_receive` callback will be invoked after processing available incoming
+        /// packets but before returning to polling the socket for additional incoming packets.
+        /// This is meant to allow the caller to bundle incoming packets into batches without
+        /// introducing delays: each time one or more packets are read from the socket there will be
+        /// a sequence of `on_receive(...)` calls for each packet, followed by a `post_receive()`
+        /// call immediately before the socket returns to waiting for additional packets.  Thus a
+        /// caller can use the `on_receive` callback to collect packets and the `post_receive`
+        /// callback to process the collected packets all at once.
         ///
         /// ev_loop must outlive this object.
-        UDPSocket(event_base* ev_loop, const Address& addr, receive_callback_t cb);
+        UDPSocket(
+                event_base* ev_loop,
+                const Address& addr,
+                receive_callback_t on_receive,
+                post_receive_callback_t post_receive = nullptr);
 
         /// Non-copyable and non-moveable
         UDPSocket(const UDPSocket& s) = delete;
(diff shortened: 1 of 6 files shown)


include/quic/udp.hpp
```diff
@@ -38,17 +38,27 @@ namespace oxen::quic
                 ;
 
         using receive_callback_t = std::function<void(const Packet& pkt)>;
+        using receive_batch_callback_t = std::function<void()>;
```
dr7ana: I have a knee-jerk reaction to suffixing with `{}_t` now and you only have yourself to blame

include/quic/udp.hpp
```diff
@@ -103,6 +113,8 @@ namespace oxen::quic
 
         event_ptr rev_ = nullptr;
         receive_callback_t receive_callback_;
+        receive_batch_callback_t receive_callback_batch_;
```
dr7ana: Maybe `batch_processing_cb` ?

src/udp.cpp
```diff
@@ -125,7 +129,16 @@ namespace oxen::quic
...
-                [](evutil_socket_t, short, void* self) { static_cast<UDPSocket*>(self)->receive(); },
+                [](evutil_socket_t, short, void* self_) {
+                    auto& self = *static_cast<UDPSocket*>(self_);
+                    self.receive();
+                    if (self.pending_receive_batch_)
+                    {
+                        self.pending_receive_batch_ = false;
+                        if (self.receive_callback_batch_)
+                            self.receive_callback_batch_();
+                    }
```
dr7ana: I'm wondering about this boolean (`self.pending_receive_batch_`). Few questions/thoughts we can chat about later:

1) I feel like there shouldn't be a case where `self.pending_receive_batch_` would be true and `self.receive_callback_batch_` would not be set? If the user doesn't provide a batch processing cb, I would guess that the batch processing logic should be avoided entirely? Though that boolean is flipped to true at line 206 `::process_packet`, implying its flipped and checked invariant of whether batch processing is "active" for this endpoint or not
2) What if we used the set-ness of the callback as an indicator of whether batch processing is "active" or "inactive" for this endpoint? Then the boolean could be a "we have a full batch" signal of sorts
3) We could wrap the callback in a lightweight struct that has two members, a callback and an int/size_t specifying the size of the batch? The size parameter could even default to something sensible at or below the loop limit of 64 packets

Follow-on... would be cool to speed-test this with different batch sizes to see what works most optimally.

Follow-on follow-on... maybe even approximate grid search cross-validation across all our configurable parameters...

tests/007-datagrams.cpp
```diff
@@ -638,4 +638,78 @@ namespace oxen::quic::test
...
+            conn->send_datagram(big_msg);
+
+        acked_first.set_value();
+
+        auto f = got_all_n_batches.get_future();
+        REQUIRE(f.wait_for(1s) == std::future_status::ready);
+        auto batch_counter_before_final = f.get();
+        REQUIRE(data_counter == 31);
+        REQUIRE(batch_counter_before_final > batches_before_flood);
+        REQUIRE(batch_counter == batch_counter_before_final + 1);
```
dr7ana: Interestingly the last `REQUIRE` failed on a few debug CI builds

jagerman:
~~I think the accept-const-string change here is actually buggy: libquic ends up storing a string_view to be used when it can actually send the packet, but we have no guarantee that the string will still be around at that point.~~

This is dropped now.

jagerman:
Keeping as draft until I've tried making this fit into Lokinet.

PR #61, opened 2023-11-03
Remove perf testing define

jagerman:
This was added so that dgram speedtest could get the dgid itself to use it to figure out the packet sequence.  The downside is that it required a compile-time change in the library.

This removes that compile-time change by changing the dgram speedtest test to mutate the last packet so that we can identify it without needing to worry about the dgid.

### DIFF
--- a/tests/dgram-speed-server.cpp
+++ b/tests/dgram-speed-server.cpp
@@ -57,9 +57,6 @@ int main(int argc, char* argv[])
 
     setup_logging(log_file, log_level);
 
-#ifdef ENABLE_PERF_TESTING
-    datagram_test_enabled = true;
-
     Network server_net{};
 
     auto server_tls = GNUTLSCreds::make(key, cert, client_cert);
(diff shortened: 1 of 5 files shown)


tests/dgram-speed-server.cpp
```diff
@@ -96,29 +90,24 @@ int main(int argc, char* argv[])
...
 
-        uint16_t raw = oxenc::load_big_to_host<uint16_t>(bstring_view{data.data(), size_t(2)}.data());
-        uint16_t _dgid = (raw >> 2);
-        uint64_t dgid = _dgid + (overflow * carry) + carry;
-
-        log::debug(test_cat, "Received datagram number: raw: {}, dgid: {}, carry: {}", raw, dgid, carry.load());
-
-        if (raw % _overflow == 0)
-            carry += 1;
+        // Subsequent packets start with a \x00 until the final one, that has first byte set to \x01.
```
tewinget:
```suggestion
        // Subsequent packets start with a \x00 until the final one; that has first byte set to \x01.
```

dr7ana (approved):

PR #64, opened 2023-11-06
Oxen push server methods + misc addenda

dr7ana:
- new static method to make gnutlscreds from one ed secret key
- connection interface getters
- btstream getters
- test code log level defaults to critical now

### DIFF
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -32,6 +32,7 @@ namespace oxen::quic
         std::string_view ep;
         std::string_view req_body;
         std::weak_ptr<BTRequestStream> return_sender;
+        ConnectionID cid;
 
       public:
         message(BTRequestStream& bp, std::string req, bool is_error = false);
@@ -61,6 +62,7 @@ namespace oxen::quic
         std::string_view body() const { return req_body; }
         std::string endpoint_str() const { return std::string{ep}; }
         std::string body_str() const { return std::string{req_body}; }
+        const ConnectionID& scid() const { return cid; }
     };
 
     struct sent_request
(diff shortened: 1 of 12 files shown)


include/quic/btstream.hpp
```diff
@@ -177,6 +187,10 @@ namespace oxen::quic
 
         void register_command(std::string endpoint, std::function<void(message)>);
 
+        const Address& local() const { return conn.local(); }
+
+        const Address& remote() const { return conn.remote(); }
+
```
jagerman, line 193: It might be more useful to make a public `connection_interface& Stream::connection()` (and then you can just `s.get_conn().local()` here instead of adding these, which seem more connection-related than stream-related).

(As a side effect that would also allow for doing things like having a command on one stream able to easily open other streams or send datagrams on the connection as a result of the first stream's message.)

include/quic/stream.hpp
```diff
@@ -119,7 +119,7 @@ namespace oxen::quic
         bool ready{false};
         int64_t _stream_id;
 
-        Connection& get_conn();
+        const Connection& get_conn() const { return conn; }
```
jagerman, line 122: We should probably have *both* of these (so that you get a const/non-const reference according to whether you have a const/non-const Stream on which you're calling it).

include/quic/gnutls_crypto.hpp
```diff
@@ -25,7 +25,9 @@ namespace oxen::quic
             unsigned int incoming,
             const gnutls_datum_t* msg)>;
 
-    constexpr size_t GNUTLS_KEY_SIZE = 32;  // for now, only supporting Ed25519 keys (32 bytes)
+    inline constexpr size_t GNUTLS_KEY_SIZE = 32;  // for now, only supporting Ed25519 keys (32 bytes)
```
jagerman, line 28: Suggested rename `GNUTLS_PUBKEY_SIZE` or `GNUTLS_PUBLIC_KEY_SIZE`.  Just `key` is ambiguous when you're in public/private keypair land, which we are with Ed25519 keys.

dr7ana: done

jagerman (approved):

PR #65, opened 2023-11-08
Address: just-port construction; is_loopback; other fixes

jagerman:
- Fixes the cmake options being too early for the is-toplevel-project detection

- Allows constructing an Address with just a port (which becomes "[::]:$PORT"), and allows removing the just-a-port constructor from opt::local_addr

- Fixes construction from empty string IP; this was supposed to be `[::]`, but I implemented it wrong and it was throwing.

- Add `is_loopback` which does what it sounds like

- Add `is_public_ip` which is like `is_public`, except that it doesn't require an addressable port.

- Add a bunch of tests for Address methods.


dr7ana (approved):
awesome

PR #66, opened 2023-11-09
Gnutls reworking

dr7ana:


include/quic/gnutls_crypto.hpp
```diff
@@ -107,9 +107,9 @@ namespace oxen::quic
     inline constexpr size_t GNUTLS_SECRET_KEY_SIZE = 64;
 
     // These bytes mean "this is a raw Ed25519 private key" in ASN.1 (or something like that)
-    inline const std::string ASN_ED25519_SEED_PREFIX = oxenc::from_hex("302e020100300506032b657004220420"sv);
+    inline const std::string_view ASN_ED25519_SEED_PREFIX{oxenc::from_hex("302e020100300506032b657004220420"sv)};
     // These bytes mean "this is a raw Ed25519 public key" in ASN.1 (or something like that)
-    inline const std::string ASN_ED25519_PUBKEY_PREFIX = oxenc::from_hex("302a300506032b6570032100"sv);
+    inline const std::string_view ASN_ED25519_PUBKEY_PREFIX{oxenc::from_hex("302a300506032b6570032100"sv)};
```
jagerman: This has to remain `std::string`: the string_view will reference the string, but the string gets destroyed immediately after the statement.

jagerman: These*

dr7ana: you're so right

src/connection.cpp
```diff
@@ -1200,6 +1213,9 @@ namespace oxen::quic
             tls_creds{context->tls_creds},
             di{*this}
     {
+        conn_established_cb = (context->conn_established_cb) ? std::move(context->conn_established_cb) : nullptr;
```
tewinget: moving this (and the next one) out of the IOContext will result in only the first connection a listener makes getting the callbacks.

dr7ana: That is entirely the point for storage server ping testing

tests/009-alpns.cpp
```diff
@@ -28,31 +27,18 @@ namespace oxen::quic::test
 
         Network test_net{};
 
-        SECTION("Do not get ALPN before negotiated")
```
tewinget, line 31: this test case still seems valid

dr7ana: yup but lets move it to 008 and rename it to "connections" then make a test case on "do not access values before connection established"

include/quic/address.hpp
```diff
@@ -249,6 +250,37 @@ namespace oxen::quic
...
+        {}
+
+        template <typename... Opt>
+        RemoteAddress(ustring_view remote_pk, Opt&&... opts) : Address{std::forward<Opt>(opts)...}, remote_pubkey{remote_pk}
+        {}
+
+        ustring_view view_remote_key() const { return remote_pubkey; }
+        const ustring& get_remote_key() const { return remote_pubkey; }
+
+        RemoteAddress(const RemoteAddress& obj) : Address{*this}, remote_pubkey{obj.remote_pubkey} { _copy_internals(obj); }
```
jagerman:
```suggestion
        RemoteAddress(const RemoteAddress& obj) : Address{obj}, remote_pubkey{obj.remote_pubkey} {}
```

Calling `Parent{*this}` in a derived constructor like this is probably undefined behaviour: at this point, the `Address` part of `RemoteAddress` hasn't been initialized yet, and so this ends up constructing `Address` by copying from itself -- even though it hasn't been initialized yet.

But we can just initialize it via copy from `obj` instead, which copies the Address portion out of `obj` into the base `Address` of `this`, and then can just initialize the RemoteAddress portions (thus without needing _copy_internals).

include/quic/address.hpp
```diff
@@ -249,6 +250,37 @@ namespace oxen::quic
...
+        RemoteAddress& operator=(const RemoteAddress& obj)
+        {
+            remote_pubkey = obj.remote_pubkey;
+            _copy_internals(obj);
```
jagerman, line 279: Here instead of `_copy_internals` it would be cleaner to invoke the parent classes' copy assignment operator, via:
```suggestion
            Address::operator=(obj);
```
to let Address worry about dealing with the Address internals.

include/quic/address.hpp
```diff
@@ -261,7 +293,7 @@ namespace oxen::quic
         ngtcp2_path _path{local, remote, nullptr};
 
       public:
-        Path() = default;
+        // Path() = default;
```
jagerman: Is making this non-default-constructible intentional?  Without this, `Path` can end up being a bit annoying to work with (i.e. I can't do something like: `Path p; if (a) p = x; else p = y;`), but perhaps that's okay with how we are using it?

include/quic/connection.hpp
```diff
@@ -150,6 +150,9 @@ namespace oxen::quic
         //		dcid: remote CID used for this connection
         //      path: network path used to reach remote client
         //      ctx: IO session dedicated for this connection context
+        //      alpns: passed directly to TLS session for handshake negotiation
+        //      remote_pk: optional parameter used by clients to verify the pubkey of the remote
+        //          endpoint during handshake negotiation
```
jagerman: Would it be correct to add `should be omitted (or nullopt) for servers` here?

include/quic/connection.hpp
```diff
@@ -150,6 +150,9 @@ namespace oxen::quic
         //		dcid: remote CID used for this connection
         //      path: network path used to reach remote client
         //      ctx: IO session dedicated for this connection context
+        //      alpns: passed directly to TLS session for handshake negotiation
```
jagerman: Are these specified in best->worst order, or worst->best?  (This probably digs down into ngtcp2 land, but would be nice to figure that out and document it here).

include/quic/connection.hpp
```diff
@@ -290,7 +306,13 @@ namespace oxen::quic
 
         dgram_interface di;
 
+        std::shared_ptr<void> user_data;
```
jagerman: Not `context`? :cry:

jagerman: (Not a serious suggestion).

include/quic/connection.hpp
```diff
@@ -290,7 +306,13 @@ namespace oxen::quic
...
+
       public:
+        void set_user_data(void* data);
+
+        std::shared_ptr<void> get_user_data() { return user_data; }
```
jagerman: ~~Should we just make the member public rather than go through a setter/getter?~~
Edit: answered out of band: no, because we want to make this abstract virtual in connection_interface with the overriding implementation here.

include/quic/endpoint.hpp
```diff
@@ -119,7 +111,7 @@ namespace oxen::quic
 
             Path _path = Path{_local, remote};
 
-            net.call([&opts..., &p, path = _path, this]() mutable {
+            net.call([&opts..., &p, path = _path, this, remote_pk = std::move(remote.get_remote_key())]() mutable {
```
jagerman: This `move` won't do anything here, because it's being passed a const reference and it won't de-const it to be able to move it.  (Technically you get back a `const Type&&` which is a fairly useless reference type in practice).

What you *could* do, however, is add a second `get_remote_key()` overload that is `&&` qualified (and change the existing one from `const` to `const &`) so that you can do `remote_pk = std::move(remote).get_remote_key()` to be able to steal the value here, i.e.

```c++
        const ustring& get_remote_key() const & { return remote_pubkey; }
        ustring&& get_remote_key() && { return std::move(remote_pubkey); }
```

jagerman: (That is assuming that `remote` is never used after this capture, of course).

include/quic/connection.hpp
tewinget: just checked, server chooses the first one on the client's list that it also supports.

include/quic/gnutls_crypto.hpp
```diff
@@ -2,20 +2,97 @@
...
 #include <variant>
 
 #include "crypto.hpp"
+#include "utils.hpp"
 
 namespace oxen::quic
 {
+    class Connection;
+
+    inline std::string translate_key_format(gnutls_x509_crt_fmt_t crt)
```
jagerman: This could be a `inline constexpr std::string_view` return type

jagerman: (The two functions below this, as well)

include/quic/address.hpp
dr7ana: yes thats better

include/quic/gnutls_crypto.hpp
```diff
@@ -25,36 +102,78 @@ namespace oxen::quic
...
+        {
+            buf = other.buf;
+            return *this;
+        }
+
+        void operator()(ustring_view data) { write(data.data(), data.size()); }
+
+        explicit operator bool() const { return not buf.empty(); }
+
+        bool operator==(const gnutls_key& other) const { return buf == other.buf; }
```
jagerman, line 128: It's pretty common to also add a `!=` operator as a matter of practice when you have a `==` operator, which is just defined by the `==` operator, i.e. just does `return !(*this == other);`

(C++20 adds the spaceship (`<=>`) operator which returns a memcmp-like -1/0/1 value, so you can define just that *one* and automatically get all of `==`, `!=`, `<`, `<=`, `>`, and `>=` provided automatically for you; in fact I think much of the STL got redefined to use that when in C++20 mode).

include/quic/address.hpp
dr7ana: Mistake, thank you (was trying out `RemoteAddress` in `Path`)

include/quic/connection.hpp
dr7ana: thanks Tom, i wasn't sure off the top of my head. added

include/quic/gnutls_crypto.hpp
```diff
@@ -25,36 +102,78 @@ namespace oxen::quic
...
         unsigned int incoming = 0;
 
         bool applies(unsigned int h, unsigned int w, unsigned int i) const
         {
-            return f && htype == h && when == w && incoming == i;
+            return cb && htype == h && when == w && incoming == i;
         }
 
-        operator bool() const { return f != nullptr; }
+        operator bool() const { return cb != nullptr; }
```
jagerman, line 151: Totally optional suggestion: this could also be `return cb;` (because std::functions are implicitly convertible to `bool`).  They do exactly the same thing, so it's just a style thing and fine either way.

include/quic/gnutls_crypto.hpp
```diff
@@ -147,6 +266,8 @@ namespace oxen::quic
             return &mem;
         }
 
+        // operator const gnutls_datum_t*() const { return &mem; }
+
```
jagerman: (just a reminder to drop this)

include/quic/connection.hpp
dr7ana: If i add virtual methods to conn interface, we can get and set from the application level. I feel like that could be useful?

include/quic/gnutls_crypto.hpp
```diff
@@ -188,47 +309,58 @@ namespace oxen::quic
         // Construct from raw Ed25519 keys
         GNUTLSCreds(std::string ed_seed, std::string ed_pubkey);
 
+      protected:
```
jagerman: superfluous (probably there was something in here before that got moved?)

include/quic/utils.hpp
```diff
@@ -147,7 +159,12 @@ namespace oxen::quic
     // Error code we send to a stream close callback if the stream's connection expires
     inline constexpr uint64_t STREAM_ERROR_CONNECTION_EXPIRED = (1ULL << 62) + 1;
 
-    // bstring_view literals baby
+    // strang literals
+    inline std::basic_string_view<unsigned char> operator""_usv(const char* __str, size_t __len) noexcept
+    {
+        return std::basic_string_view<unsigned char>(reinterpret_cast<const unsigned char*>(__str), __len);
```
jagerman: This could use the `ustring_view` typedef for convenience.  (And a `bstring_view` for the one below).

include/quic/connection.hpp
jagerman: Copying from out-of-band discussion:

```C++
In connection_interface:
template <typename T, typename... Args>
void make_user_data(Args&&... args) {
    get_user_data() = std::make_shared<T>(std::forward(args)...);
}

virtual std::shared_ptr<void>& get_user_data() = 0;

template <typename T, typename = std::enable_if_t<!std::is_void<T>>>
std::shared_ptr<T> get_user_data() {
    return std::static_pointer_cast<T>(get_user_data());
}
```

Or perhaps:
```C++
protected:

virtual std::shared_ptr<void>& user_data_ptr() = 0;

public:

template <typename T, typename... Args>
void make_user_data(Args&&... args) {
    user_data_ptr() = std::make_shared<T>(std::forward(args)...);
}

public:
template <typename T, typename = std::enable_if_t<!std::is_void<T>>>
std::shared_ptr<T> get_user_data() {
    return std::static_pointer_cast<T>(user_data_ptr());
}
```

And then in Connection:
```
std::shared_ptr<void>& user_data_ptr() override { return user_data; }
```

include/quic/opt.hpp
```diff
@@ -13,9 +13,9 @@ namespace oxen::quic::opt
         using Address::Address;
     };
 
-    struct remote_addr : public Address
+    struct remote_addr : public RemoteAddress
     {
-        using Address::Address;
+        using RemoteAddress::RemoteAddress;
     };
```
jagerman: I think we could get rid of these two since I think they are never really optional values, and always the mandatory first argument for `listen` or `connect`.  *Perhaps* we could change that mandatory first argument to be a `opt::{local,remote}_addr`, but I'd be just as fine with leaving it as Address/RemoteAddress and dropping these (since it isn't like the other `opt::` values that go through the almost-named-argument argument handling).

src/connection.cpp
```diff
@@ -1200,6 +1213,9 @@ namespace oxen::quic
...
+        conn_closed_cb = (context->conn_closed_cb) ? std::move(context->conn_closed_cb) : nullptr;
+
```
jagerman, line 1204: Discussed offline, but just dropping a reminder comment here: these two callbacks are a little bit weird since they are dual-purposed for both incoming and outgoing; I think we should split them up with separate callbacks specifiable for incoming and outgoing connection establishment/closing.

src/connection.cpp
```diff
@@ -1284,11 +1309,12 @@ namespace oxen::quic
             std::shared_ptr<IOContext> ctx,
             const std::vector<std::string>& alpns,
             std::chrono::nanoseconds handshake_timeout,
+            std::optional<ustring> remote_pk,
             ngtcp2_pkt_hd* hdr)
     {
         log::trace(log_cat, "{} called", __PRETTY_FUNCTION__);
         std::shared_ptr<Connection> conn{
-                new Connection{ep, scid, dcid, path, std::move(ctx), alpns, handshake_timeout, hdr}};
+                new Connection{ep, scid, dcid, path, std::move(ctx), alpns, handshake_timeout, remote_pk, hdr}};
```
jagerman: Can usefully `std::move` the `remote_pk` value here (I think everything else here not already being moved is trivially copyable).

src/gnutls_creds.cpp
```diff
@@ -0,0 +1,277 @@
...
+    {
+        // NOTE: I put this here (rather than just the raw_pk constructor) to test the server requesting certs every time,
+        // but not requiring them. By doing the check here, we can ensure that requesting is only done by servers, while
+        // retrieval is only done by clients
+        if (using_raw_pk)
+        {
+            // if (is_client)
+            //     gnutls_certificate_set_retrieve_function2(cred, cert_retrieve_callback_gnutls);
+            // else
+            gnutls_certificate_set_verify_function(cred, cert_verify_callback_gnutls);
```
jagerman: This feels a little akward because we'll be setting this on `creds` again and again for every session we create.  Reading the comment above, I'm assuming this is just for the test suite, so I'm wondering if we can find a way to avoid the awkwardness?

src/gnutls_session.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+            return tls_session->do_tls_callback(session, htype, when, incoming, msg);
+        }
+
+        int gnutls_post_handshake(gnutls_session_t session)
+        {
+            auto* tls_session = get_session_from_gnutls(session);
+            // Same notes on the assert as in gnutls_callback_wrapper above
+            assert(tls_session->get_session() == session);
+            (void)tls_session;
```
jagerman, line 28: An alternative to these three lines that won't hit the unused variable warning is to inline it all inside the assert (sometimes that gets hairy, but here seems fairly innocuous):
```c++
assert(get_session_from_gnutls(session)->get_session() == session);
```

tests/001-handshake.cpp
```diff
@@ -108,68 +109,116 @@ namespace oxen::quic::test
...
-            opt::local_addr client_remote{"127.0.0.1"s, server_endpoint->local().port()};
+            auto client_closed = callback_waiter{[&client_error](connection_interface&, uint64_t) { client_error = 1000; }};
+            opt::remote_addr client_remote{defaults::CLIENT_PUBKEY, "127.0.0.1"s, server_endpoint->local().port()};
 
-            auto client_endpoint = test_net.endpoint(client_local);
-            REQUIRE_NOTHROW(client_endpoint->connect(client_remote, client_tls));
+            auto client_endpoint = test_net.endpoint(client_local, client_established, client_closed);
+            auto client_ci = client_endpoint->connect(client_remote, client_tls);
+
+            REQUIRE(not client_established.wait());
```
jagerman, line 149: Shouldn't this be a `REQUIRE(client_closed.wait())` since that's what we expect to trigger in this case?

tests/001-handshake.cpp
```diff
@@ -108,68 +109,116 @@ namespace oxen::quic::test
...
+        REQUIRE(server_endpoint->listen(server_tls));
 
-            auto server_endpoint = test_net.endpoint(server_local);
-            REQUIRE(server_endpoint->listen(server_tls));
+        SECTION("Endpoint::listen() + Endpoint::Connect() - Incorrect pubkey in remote")
+        {
+            uint64_t client_error{0};
 
-            opt::local_addr client_remote{"127.0.0.1"s, server_endpoint->local().port()};
+            auto client_closed = callback_waiter{[&client_error](connection_interface&, uint64_t) { client_error = 1000; }};
```
jagerman, line 143: Is there a particularly errcode value (passed in the second argument) for this mismatched-key failure?  Or to put it another way, how could an application distinguish between a bad pubkey, and, say, a timeout?

tests/utils.hpp
```diff
@@ -17,6 +18,26 @@ namespace oxen::quic
 
     inline auto test_cat = oxen::log::Cat("test");
 
+    namespace test::defaults
+    {
+        inline const std::string CLIENT_SEED{
+                oxenc::from_hex("468e7ed2cd914ca44568e7189245c7b8e5488404fc88a4019c73b51d9dbc48a5"sv)};
```
jagerman: oxenc provides a `_hex` we could use here:
```C++
using namespace oxenc::literals;

// ...

inline const std::string CLIENT_SEED = "468e7ed2cd914ca44568e7189245c7b8e5488404fc88a4019c73b51d9dbc48a5"_hex;
```

tests/utils.hpp
```diff
@@ -17,6 +18,26 @@ namespace oxen::quic
...
+        inline const std::string CLIENT_PUBKEY{
+                oxenc::from_hex("626136fe40c8860ee5bdc57fd9f15a03ef6777bb9237c18fc4d7ef2aacfe4f88"sv)};
+        inline const std::string SERVER_SEED{
+                oxenc::from_hex("fefbb50cdd4cde3be0ae75042c44ff42b026def4fd6be4fb1dc6e81ea0480c9b"sv)};
+        inline const std::string SERVER_PUBKEY{
+                oxenc::from_hex("d580d5c68937095ea997f6a88f07a86cdd26dfa0d7d268e80ea9bbb5f3ca0304"sv)};
```
jagerman: I'm thinking that it might be nice to have this randomize the keypairs on each initialization; with libsodium that's trivial, but I'm not sure if gnutls gives us something nice to do that or not.

jagerman: So we could use `nettle` for this (we already have or build it, as it's a dependency of gnutls), but I'm not sure whether it's worth worrying about right now.  The fixed pregenerated keys are probably okay.

jagerman (commented):
Done first review pass, comments left.

include/quic/connection.hpp
dr7ana: Thanks for copying this here!!

include/quic/gnutls_crypto.hpp
dr7ana: Nice thanks added

include/quic/gnutls_crypto.hpp
dr7ana: That's a good point didn't even cross my mind

include/quic/gnutls_crypto.hpp
dr7ana: just protecting my privates ya know

include/quic/utils.hpp
dr7ana: Nice they were even at the top, changed

include/quic/opt.hpp
dr7ana: I lean towards using the `Address` and `RemoteAddress` types explicitly. I think the idea of using the opt structs as packs is a better use for it anyways

src/connection.cpp
dr7ana: Summarizing off-band discussion to ensure I'm on the same page:

We have two asymmetries in our connection open/close callbacks:
1) Callbacks for outbounds (passed to `Endpoint::connect(...)`) are to be used once, but callbacks for inbounds (passed to `Endpoint::listen(...)`) should be used for all inbound connections to that address
2) Our callback types should be differentiated between inbound/outbound callbacks, which may be another use-case for opt?

I can easily do bullet 1 in this PR; bullet 2 is a good stand-alone PR?

src/gnutls_creds.cpp
dr7ana: Moved into the creds constructor so it's done once at creation

tests/001-handshake.cpp
dr7ana: A normal (not error-caused) closure would fire with an error code of 0, so any non-zero error code is a problem.

Currently, a pubkey that could not be validated is still an allowed connection, so it wouldn't fire the connection closed callback. Rather, it would fire the connection established callback, and the user checks `connection_interface.is_validated()`

tests/utils.hpp
dr7ana: Good idea. I messed with the `gnutls_rnd` function a bit but it was fruitless in this regard. Gnutls also has specific methods for generating ed keys, but it has to be directly into their abstract types like `gnutls_privkey_t`

include/quic/connection.hpp
dr7ana: Follow-up on out of band discussion:
We want to differentiate between `get_user_data` and `take_user_data` right? Since the virtual method returns a `shared_ptr` reference, calling the templated method would increase the use count to 2, making it impossible for the user to destroy the data inside `Connection` by calling `::reset()` on their `shared_ptr`.

I know the user could call `get_user_data<T>()` to steal it, but maybe we have two methods?
- `view_user_data` to access it
- `get_user_data` to take control of it

src/gnutls_session.cpp
dr7ana: this actually still flipped an unused variable error for the parameter session

include/quic/gnutls_crypto.hpp
dr7ana: Narrator: "No they couldn't..."

[here](https://ci.oxen.rocks/dr7ana/libquic/290/2/3)

jagerman (approved):

PR #69, opened 2023-11-24
Optional options

jagerman:
Certain options in the current interface in awkward in contexts such as Python where whether you want an option is dependent on some runtime value: such as being passed a `datagram=True` keyword argument.

Currently the only way we could do it would be to have two separate paths for each non-nullable option, i.e. something like:

```C++
    std::shared_ptr<Endpoint> e;
    if (want_datagram) {
        e = net.endpoint(addr, opt::whatever{},
            opt::datagram_enabled{});
    } else {
        e = net.endpoint(addr, opt::whatever{});
    }
```
because there's no way to make a `opt::datagram_enabled` that means *don't* enable datagrams.

While we *could* add such a feature to datagram_enabled and any other using-this-turns-it-on option, it seems a lot simpler to add a generic `std::optional<T>` handle to option parsing so that you can pass a `nullopt` anywhere you don't want to pass an option, i.e. the above can become:

```C++
    auto e = net.endpoint(addr, opt::whatever{},
        want_datagrams ? std::make_optional<opt::datagram_enabled>() : std::nullopt);
```
and similarly for any other of the templated option types.

### DIFF
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -112,6 +112,13 @@ namespace oxen::quic
       private:
         void handle_req_opts(std::function<void(message)> func) { cb = std::move(func); }
         void handle_req_opts(std::chrono::milliseconds exp) { timeout = exp; }
+
+        template <typename Opt>
+        void handle_req_opts(std::optional<Opt> option)
+        {
+            if (option)
+                handle_req_opts(std::move(*option));
+        }
     };
 
     class BTRequestStream : public Stream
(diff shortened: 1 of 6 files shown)


dr7ana:
Maybe I'm missing something, but I don't see how going from this...
```C++
if (want_datagram) {
    e = net.endpoint(addr, opt::whatever{},
        opt::datagram_enabled{});
} else {
    e = net.endpoint(addr, opt::whatever{});
}
```
to this...
```C++
auto e = net.endpoint(addr, opt::whatever{},
    want_datagrams ? std::make_optional<opt::datagram_enabled>() : std::nullopt);
```
is an optimization in anything except LOC? We have defaulted values in the Connection constructor, why can't we just default values in the endpoint constructor similarly?

Just looking at the resulting network constructor in pyoxquic with four ternaries in the constructor feels super weird to me

dr7ana (approved):
Looks good, we can cherry pick #70 into this?

PR #71, opened 2023-11-30
Various changes from pybinding

jagerman:
- Change `listen` return to `void`: it only ever throws, or returns true, so convert it to void with a void promise so now it either throws on failure, or doesn't throw on success.
- Moved the direction methods up to `connection_interface` so that a public caller (like pybind) can get them.
- Removed a dubious-looking double-enable_shared_from_this inheritance from Connection

### DIFF
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -48,7 +48,6 @@ namespace oxen::quic
         void respond(bstring_view body, bool error = false);
         void respond(std::string_view body, bool error = false) { respond(convert_sv<std::byte>(body), error); }
 
-        bool timed_out{false};
         bool is_error{false};
 
         //  To be used to determine if the message was a result of an error as such:
@@ -61,7 +60,7 @@ namespace oxen::quic
         //      if (m)
         //      { // success logic }
         //  }
-        operator bool() const { return not timed_out && not is_error; }
+        operator bool() const { return not is_error; }
 
         template <typename Char = char, typename = std::enable_if_t<sizeof(Char) == 1>>
         std::basic_string_view<Char> view() const
@@ -70,7 +69,10 @@ namespace oxen::quic
         }
 
         int64_t rid() const { return req_id; }
-        std::string_view type() const { return {reinterpret_cast<const char*>(data.data()) + req_type.first, req_type.second}; }
+        std::string_view type() const
+        {
+            return {reinterpret_cast<const char*>(data.data()) + req_type.first, req_type.second};
+        }
         std::string_view endpoint() const { return {reinterpret_cast<const char*>(data.data()) + ep.first, ep.second}; }
         std::string endpoint_str() const { return std::string{endpoint()}; }
 
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -29,7 +29,9 @@ namespace oxen::quic
 
       private:
         int64_t req_id;
-        bstring data;
+        bstring data;  // Wrapped in a shared pointer because: a) we store views into this and need
+                       // them to stay valid across moves; and b) we want message to be copyable so
+                       // that it can be captured in lambda args.
         std::string_view req_type;
         std::string_view ep;
         std::string_view req_body;
@@ -39,6 +41,11 @@ namespace oxen::quic
         message(BTRequestStream& bp, bstring req, bool is_error = false);
 
       public:
+        message(message&& m);
+        message(const message& m);
+        message& operator=(message&& m);
+        message& operator=(const message& m);
+
         void respond(bstring_view body, bool error = false);
         void respond(std::string_view body, bool error = false) { respond(convert_sv<std::byte>(body), error); }
 
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -66,10 +66,10 @@ namespace oxen::quic
 
         std::shared_ptr<BTRequestStream> stream() const
         {
-            if (return_sender.expired())
-                throw std::runtime_error{"Cannot access expired pointer to BT stream!"};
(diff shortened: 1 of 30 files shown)


tewinget (approved):

tewinget (approved):

src/stream.cpp
```diff
@@ -161,7 +161,7 @@ namespace oxen::quic
     void Stream::send_impl(bstring_view data, std::shared_ptr<void> keep_alive)
     {
         if (data.empty())
-            throw std::invalid_argument{"Cannot send empty byte string"};
```
dr7ana, line 164: I honestly kinda like that this throws on an empty send? It immediately tells us that something is wrong in whatever application logic is trying to feed data to libquic

Though I've been off in refactor land for a long time, def gonna pick your brain off band on this

jagerman: > I honestly kinda like that this throws on an empty send? It immediately tells us that something is wrong in whatever application logic is trying to feed data to libquic

The thing is, it only *might* be a bug in application logic, but by throwing here we're forcing it to be an application bug.  It's kind of like `std::cout << "";` -- it does nothing, but isn't necessarily a bug.

jagerman: Case in point: my echo-server broke without this because it reads a line of input, then sends it, and broke when I hit enter without typing anything.  It feels weird for that to be an invalid use of `send()` and for me to have to make sure that anywhere I call send() I've already checked for non-emptiness.

dr7ana: Shit. Valid.

include/quic/datagram.hpp
```diff
@@ -47,7 +47,7 @@ namespace oxen::quic
     };
 
     // IO callbacks
-    using dgram_data_callback = std::function<void(dgram_interface&, bstring)>;
+    using dgram_data_callback = std::function<void(std::shared_ptr<dgram_interface>, bstring)>;
```
jagerman: We do not want to pass a shared_ptr here (or for the stream data): it will incur the overhead of instantiating an additional shared_ptr instance (which requires an atomic reference count increment on construction and decrement on destruction) on every data callback, even if the caller does not care about or need a shared ptr.  By passing it as a reference we avoid that overhead and only incur if if the caller needs it (by calling `shared_from_this()` on the reference).

The relatively infrequent calls (stream open, etc.) could be passed by shared_ptr, but I'm not sure it's worthwhile to be inconsistent (references for data callbacks, shared_ptr for other things).

include/quic/connection.hpp
```diff
@@ -90,6 +91,12 @@ namespace oxen::quic
             }));
         }
 
+        template <typename StreamT = Stream, std::enable_if_t<std::is_base_of_v<Stream, StreamT>, int> = 0>
+        std::shared_ptr<StreamT> get_stream(int64_t id)
+        {
+            return std::static_pointer_cast<StreamT>(get_stream_impl(id));
```
jagerman: This needs to be a dynamic_pointer_cast if you can't guarantee that the pointer is actually a StreamT

include/quic/btstream.hpp
```diff
@@ -29,23 +29,22 @@ namespace oxen::quic
...
-                       // that it can be captured in lambda args.
-        std::string_view req_type;
-        std::string_view ep;
-        std::string_view req_body;
+        bstring data;
+
+        // We keep the locations of variables fields as relative positions inside `data` *rather*
+        // than using std::string_view members because the string_views are more difficult to
+        // maintain when the object gets moved or copied.
+        using substr_location = std::pair<std::ptrdiff_t, std::size_t>;
```
dr7ana, line 37: This would be a really cool idea for a PR to oxen-encoding, basically extracting a map or some referential structure indicating the indexes of each data field in the bt-dict or list

tests/002-send-receive.cpp
```diff
@@ -269,6 +351,206 @@ namespace oxen::quic::test
...
+            std::shared_ptr<BTRequestStream> client_bp = conn_interface->get_new_stream<BTRequestStream>();
+
+            for (int i = 0; i < num_requests; i++)
+            {
+                client_bp->command("test_endpoint"s, req_msg, client_reply_handler);
+            }
+
+            CHECK(done.wait_for(5s) == std::future_status::ready);
+            std::lock_guard lock{mut};
+            CHECK(good_responses == num_requests);
```
dr7ana, line 505: Okay so I dug into why this is failing literally all the CI builds. I separated the test cases to look into the one that was failing specifically, pushed it to [my branch](https://github.com/dr7ana/libquic/blob/f20358f345e463c773c41452972a7a7764ca0f6f/tests/002-send-receive.cpp#L432) to keep testnet tracking the pr void-listen

[This is the offending line](https://github.com/dr7ana/libquic/blob/f20358f345e463c773c41452972a7a7764ca0f6f/src/connection.cpp#L910) that slows down the test to the point of failure, and downgrading it's log-level to trace makes the test work as you wrote it.

Though what's interesting is if you leave that line as a debug log, put [this line](https://github.com/dr7ana/libquic/blob/f20358f345e463c773c41452972a7a7764ca0f6f/tests/002-send-receive.cpp#L513) instead of the `done.wait_for(5s)` call, and run the test suit with debug or trace logging, it loops forever (both with and without the mutexes)

src/btstream.cpp
```diff
@@ -25,6 +25,8 @@ namespace oxen::quic
             ep = get_location(data, btlc.consume_string_view());
 
         req_body = get_location(data, btlc.consume_string_view());
+
+        btlc.finish();
```
dr7ana, line 29: I dunno about this change i really need to think about it.

@tewinget thoughts?

PR #72, opened 2023-12-14
Stream open/queuing redux

jagerman:
This PR addresses some issues with btrequest error/timeout flags; and some issues with how streams get constructed and queued with some conflicts when trying to mix queue_stream, get_new_stream, and a stream constructor callback.

(This is based on #71; only the last two commits here are relevant to this PR)

As for the stream changes:

There was a confusing behaviour with `queue_stream` and `get_new_stream` that, even if given an explicit stream type and arguments, those arguments and type would be ignored if there was a stream constructor callback and so you really had to go *only* stream constructor callback *or* get_new_stream/queue_stream because mixing them would break in unexpected ways.

This addresses the issue (along with related changes along the way):

- rename `queue_stream` to `queue_incoming_stream` to better reflect what it does.
- rename `get_new_stream` to `open_stream` to better reflect what it does.
- split `queue_incoming_stream` and `open_stream` into two versions:
    - an untemplated version that takes no arguments and always just constructs a default stream (firing the stream constructor callback, if set; making a basic Stream otherwise).
    - a templated version that always bypasses the stream constructor callback and always creates the requested type with the given arguments.
- Added a `maybe_stream` alternative to `get_stream` that returns nullptr if the stream id doesn't exist, and throws if the stream exists but isn't compatible with the given templated type (if given a Stream subclass to cast to).
- Changed `get_stream` to throw a different exception (invalid_argument) if the stream exists but has an incompatible type, and changed its stream-does-not-exist exception to a std::out_of_range.
- added a virtual `on_ready()` function to Stream that subclasses can override if they want to do something when the stream becomes ready.
- various test cases for the new stream queuing/opening/retrieving that adds various additional tests for edge cases and mixed queuing/constructor callback usage.

### DIFF
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -48,7 +48,6 @@ namespace oxen::quic
         void respond(bstring_view body, bool error = false);
         void respond(std::string_view body, bool error = false) { respond(convert_sv<std::byte>(body), error); }
 
-        bool timed_out{false};
         bool is_error{false};
 
         //  To be used to determine if the message was a result of an error as such:
@@ -61,7 +60,7 @@ namespace oxen::quic
         //      if (m)
         //      { // success logic }
         //  }
-        operator bool() const { return not timed_out && not is_error; }
+        operator bool() const { return not is_error; }
 
         template <typename Char = char, typename = std::enable_if_t<sizeof(Char) == 1>>
         std::basic_string_view<Char> view() const
@@ -70,7 +69,10 @@ namespace oxen::quic
         }
 
         int64_t rid() const { return req_id; }
-        std::string_view type() const { return {reinterpret_cast<const char*>(data.data()) + req_type.first, req_type.second}; }
+        std::string_view type() const
+        {
+            return {reinterpret_cast<const char*>(data.data()) + req_type.first, req_type.second};
+        }
         std::string_view endpoint() const { return {reinterpret_cast<const char*>(data.data()) + ep.first, ep.second}; }
         std::string endpoint_str() const { return std::string{endpoint()}; }
 
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -29,7 +29,9 @@ namespace oxen::quic
 
       private:
         int64_t req_id;
-        bstring data;
+        bstring data;  // Wrapped in a shared pointer because: a) we store views into this and need
+                       // them to stay valid across moves; and b) we want message to be copyable so
+                       // that it can be captured in lambda args.
         std::string_view req_type;
         std::string_view ep;
         std::string_view req_body;
@@ -39,6 +41,11 @@ namespace oxen::quic
         message(BTRequestStream& bp, bstring req, bool is_error = false);
 
       public:
+        message(message&& m);
+        message(const message& m);
+        message& operator=(message&& m);
+        message& operator=(const message& m);
+
         void respond(bstring_view body, bool error = false);
         void respond(std::string_view body, bool error = false) { respond(convert_sv<std::byte>(body), error); }
 
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -66,10 +66,10 @@ namespace oxen::quic
 
         std::shared_ptr<BTRequestStream> stream() const
         {
-            if (return_sender.expired())
-                throw std::runtime_error{"Cannot access expired pointer to BT stream!"};
+            if (auto ptr = return_sender.lock())
+                return ptr;
 
-            return return_sender.lock();
+            throw std::runtime_error{"Cannot access expired pointer to BT stream!"};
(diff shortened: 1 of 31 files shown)


include/quic/connection.hpp
```diff
@@ -55,48 +56,97 @@ namespace oxen::quic
...
+        /// comments in the templated version of the method, above, for details about the readiness
+        /// of the returned stream.
+        std::shared_ptr<Stream> open_stream();
+
+        /// Returns a stream object for the stream with the given id, if the stream exists (and, if
+        /// StreamT is specified, is of the given Stream subclass).  Returns nullptr if the id is
+        /// not currently an open stream; throws std::invalid_argument if the stream exists but is
+        /// not an instance of the given StreamT type.
+        template <typename StreamT = Stream, std::enable_if_t<std::is_base_of_v<Stream, StreamT>, int> = 0>
+        std::shared_ptr<StreamT> maybe_stream(int64_t id)
```
dr7ana, line 126: thanks for adding this!

include/quic/connection.hpp
```diff
@@ -55,48 +56,97 @@ namespace oxen::quic
...
+        {
+            auto s = get_stream_impl(id);
+            if (!s)
+                return nullptr;
+            if constexpr (!std::is_same_v<StreamT, Stream>) {
+                if (auto st = std::dynamic_pointer_cast<StreamT>(std::move(s)))
+                    return st;
+                throw std::invalid_argument{"Stream ID {} is not an instance of the requested Steam subclass"_format(id)};
```
dr7ana: Small typo ("Steam")

include/quic/connection.hpp
```diff
@@ -55,48 +56,97 @@ namespace oxen::quic
     class connection_interface : public std::enable_shared_from_this<connection_interface>
     {
       protected:
-        virtual std::shared_ptr<Stream> queue_stream_impl(
+        virtual std::shared_ptr<Stream> queue_incoming_stream_impl(
                 std::function<std::shared_ptr<Stream>(Connection& c, Endpoint& e)> make_stream) = 0;
-        virtual std::shared_ptr<Stream> get_new_stream_impl(
+        virtual std::shared_ptr<Stream> open_stream_impl(
```
dr7ana, line 61: I think this re-naming is a lot clearer on intent and purpose

include/quic/btstream.hpp
```diff
@@ -36,7 +36,7 @@ namespace oxen::quic
         std::weak_ptr<BTRequestStream> return_sender;
         ConnectionID cid;
 
-        message(BTRequestStream& bp, bstring req, bool is_error = false);
+        message(BTRequestStream& bp, bstring req, bool is_timeout = false);
 
       public:
         void respond(bstring_view body, bool error = false);
```
dr7ana, line 50: Discussion point:
Currently our response types are "errors" ("E") and "responses" ("R"), but we could introduce a "timeout" ("T") option to actually make a useful differentiation between the two in the application code. Otherwise, we'd have to use some standardized "timeout response" that could be parsed by the library code

src/connection.cpp
```diff
@@ -373,25 +373,31 @@ namespace oxen::quic
...
 
-    std::shared_ptr<Stream> Connection::queue_stream_impl(
+    std::shared_ptr<Stream> Connection::queue_incoming_stream_impl(
             std::function<std::shared_ptr<Stream>(Connection& c, Endpoint& e)> make_stream)
     {
         return _endpoint.call_get([this, &make_stream]() {
-            auto stream = construct_stream(make_stream);
+            std::shared_ptr<Stream> stream;
+            if (make_stream)
+                stream = make_stream(*this, _endpoint);
```
dr7ana, line 396: Why do you only pass the optional ID to the make_stream in `Connection::stream_opened`? This completely eliminates the use case of queueing streams to be opened in the client's on_conn_open callback, for example

dr7ana:
I added a few test cases: 
The first test case extracts streams created in the server stream constructor and confirming ID's and such. If using the stream constructor, the server cannot use `maybe_stream` to get the btstream until the client has sent a request down

The other was an issue i noticed in stream queueing for inbound connections. If a server queue's a btreq stream to be opened in its connection open constructor (like in lokinet), it cannot use the `::maybe_stream` method to get it until the client sends something down that stream, upon which it is moved from `stream_queue` to `streams`. I made a quick change to `Connection::get_stream_impl` to check `stream_queue` after checking `streams,` which passed the test case.

Though they went from all failing from 002 (from the connection line that prints the huge requests) to failing in the 004 test cases prior to the ones i added i think... i haven't been able to reproduce locally either which is weird

include/quic/btstream.hpp
jagerman: Timeout is not a response; it happens because we didn't get *any* response to the request within a certain amount of time.

PR #77, opened 2024-01-07
BT stream request queue

dr7ana:
- no substantial changes to the internal code are necessary; a few methods to return queue size are added
- new test case in 004 displaying the ability of outbound streams to have send queues, supplanting the need for any in the application layer

### DIFF
--- a/include/quic/datagram.hpp
+++ b/include/quic/datagram.hpp
@@ -72,7 +72,6 @@ namespace oxen::quic
         virtual bool is_empty() const = 0;
         virtual std::shared_ptr<Stream> get_stream() = 0;
         virtual std::vector<ngtcp2_vec> pending() = 0;
-        virtual size_t num_pending() const = 0;
         virtual prepared_datagram pending_datagram(bool) = 0;
         virtual int64_t stream_id() const = 0;
         virtual bool is_closing() const = 0;
@@ -194,8 +193,6 @@ namespace oxen::quic
         buffer_que send_buffer;
 
         bool is_empty() const override { return send_buffer.empty(); }
-
-        size_t num_pending() const override { return send_buffer.size(); }
 
         prepared_datagram pending_datagram(bool r) override;
 
--- a/include/quic/datagram.hpp
+++ b/include/quic/datagram.hpp
@@ -72,6 +72,7 @@ namespace oxen::quic
         virtual bool is_empty() const = 0;
         virtual std::shared_ptr<Stream> get_stream() = 0;
         virtual std::vector<ngtcp2_vec> pending() = 0;
+        virtual size_t num_pending() const = 0;
         virtual prepared_datagram pending_datagram(bool) = 0;
         virtual int64_t stream_id() const = 0;
         virtual bool is_closing() const = 0;
@@ -193,6 +194,8 @@ namespace oxen::quic
         buffer_que send_buffer;
 
         bool is_empty() const override { return send_buffer.empty(); }
+
+        size_t num_pending() const override { return send_buffer.size(); }
 
         prepared_datagram pending_datagram(bool r) override;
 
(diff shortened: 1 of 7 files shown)


include/quic/stream.hpp
```diff
@@ -82,7 +81,7 @@ namespace oxen::quic
         }
 
         // Called immediately after set_ready so that a subclass can do thing as soon as the stream
-        // becomes ready.  The default does nothing.
+        // becomes ready. \The default does nothing.
```
jagerman: \

include/quic/datagram.hpp
```diff
@@ -72,6 +72,7 @@ namespace oxen::quic
         virtual bool is_empty() const = 0;
         virtual std::shared_ptr<Stream> get_stream() = 0;
         virtual std::vector<ngtcp2_vec> pending() = 0;
+        virtual size_t num_pending() const = 0;
```
jagerman: I don't think this `num_pending()` should be in the base class, because I don't think it has a consistent meaning (or even a useful meaning at all in some cases) across different subclasses.

- for datagrams this is unsent datagrams.  This seems pretty unambiguous and is possibly useful thing to be able to query, but is quite specific to datagrams.

- for regular Streams this is the number of internal buffers that have not been fully acked by the other side.  This one doesn't seem meaningful on a plain Stream, though, because it's a byte-oriented stream of bytes, and this exposes an internal implementation detail of Stream that makes it look discrete, when Streams really aren't discrete and shouldn't be treated as such.  We already have `size()`, `unsent()` and `unacked()` on streams (each returning byte counts—i.e. count of the Stream's discrete unit) which are really the only things we should expose for streams.  And because we have all of those three a `num_pending()` seems like it's just exposing an implementation detail that we probably shouldn't expose.  (In particular, it would somewhat lock us into the current behaviour of how Stream handles data buffers, because anything that changes the data buffer handling would potentially break the public API `num_pending()` value).

- for bt request streams, as a side effect of the way we push data into the base Stream, the current base Stream implementation counting current user buffers ends up being the sum of: the number of possibly sent, but not-fully-acked outgoing requests/commands; the number of possibly sent but not-fully-acked outgoing responses; and the number of possibly sent, but not-fully-acked outgoing errors that we've queue over the stream.  None of these feel like what we want: if I have a bt request stream object and count `num_pending()` then shouldn't I expect the number of outstanding requests, rather than :point_up_2: that?

  In fact, using the stream internal buffers seems like it could be quite misleading: suppose I initiate a request that the other end puts in a job queue but takes a second to come back.  The stream *buffers* will be completely empty long before I get the response (since the request has been sent and ACKed in the quic stream), and so `num_pending()` will give me back 0, even though I still very much have a pending request.
  
  For a bt request stream I think that the number of unanswered (not unacked) requests is the thing that I, as a library user, would care about—and I think is the intention, from reading the example added in this PR.  That is going to be just `sent_reqs.size()`, and not dependent on the stream internal buffers.

dr7ana: I needed something for the test code to indicate how many packets were on deck prior to the connection being ready to transmit. Calling `sent_reqs.size()` doesn't quantify the same thing on the BT stream, do you have a better idea that satisfies the test code need? This was a quick fix for that, open to changing it to whatever

jagerman: Wouldn't `sent_reqs.size()` do what you need?

dr7ana: No, because they're not sent at that point. The whole point of doing it this way was to get all 5 messages queued up in `user_buffers,` not `sent_reqs`. I'll  take the virtual method for ::num_pending() out of IOChannel and the other types, and keep it just in BTReqStream to fix this then?

jagerman: I don't see why a `num_pending()` call is needed in the test.  You're in the key verify function where this gets used, so the stream isn't ready yet, and so a check against `sent_reqs.size() == 5`  and `unsent() > 0` would tell you that 1) you have queued requests, and 2) that you have unsent, queued stream data.  Why does this test need to test specifically that there are exactly 5 internal buffers in use in the stream?

dr7ana: Resolving this after our off-band discussions. Will modify as we discussed, and likely loop it into #78; should I close this PR?

dr7ana:
Closing this for #78

PR #78, opened 2024-01-08
Migration/resumption

dr7ana:


include/quic/address.hpp
```diff
@@ -303,13 +314,19 @@ namespace oxen::quic
...
+
       public:
         Address local;
         Address remote;
 
       private:
         ngtcp2_path _path{local, remote, nullptr};
 
+        void set_new_remote(const ngtcp2_addr& new_remote, uint16_t p);
+        void set_new_local(const ngtcp2_addr& new_local, uint16_t p);
```
jagerman: It feels odd for these to take a port, since there is already a port inside the ngtcp2_addr.  Is there a reason we can't use the one inside `new_remote`/`new_local`?

include/quic/address.hpp
```diff
@@ -25,6 +27,13 @@ namespace oxen::quic
         // Default constructor or single-port constructor yields [::]:port (or [::]:0 if port omitted)
         explicit Address(uint16_t port = 0) : Address{"", port} {}
 
+        Address(const sockaddr* s, socklen_t n, uint16_t p)
+        {
+            std::memmove(&_sock_addr, s, n);
+            _addr.addrlen = n;
+            set_port(p);
+        }
```
jagerman: This constructor feels weird too because port is passed twice (once inside `s`, and again in `p`).

include/quic/connection_ids.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+namespace oxen::quic
+{
+    // Unique reference ID for each individual connection. QUIC allows for a
+    // connection to be reached at any of n number of connection ID's (see above
+    // constexpr) at a time. As a result, we key each of those QUIC CID's to the
+    // unique reference ID, to which the connection shared_ptr is keyed in the
+    // master map container
+    struct alignas(uint64_t) ReferenceID final
+    {
```
jagerman, line 21: The name here felt a little bit odd to use in storage server, because I have a "oxen::quic::ReferenceID" but it isn't clear from the name what it is a reference to (Endpoint? Connection? Stream?).  Perhaps renaming to "ConnRefID" or something along those lines to indicate that it is specific to connections would be useful?

src/gnutls_session.cpp
```diff
@@ -1,86 +1,100 @@
...
 
     GNUTLSSession::GNUTLSSession(
-            GNUTLSCreds& creds, bool is_client, const std::vector<ustring>& alpns, std::optional<gnutls_key> expected_key) :
-            creds{creds}, is_client{is_client}
+            GNUTLSCreds& creds, Connection& c, const std::vector<ustring>& alpns, std::optional<gnutls_key> expected_key) :
+            creds{creds}, session_ticket_key{}, is_client{c.is_outbound()}
     {
         log::trace(log_cat, "Entered {}", __PRETTY_FUNCTION__);
 
+        if (not is_client)
```
jagerman, line 83:
```suggestion
        if (false && not is_client)  // TODO: we need this once we start supporting 0-RTT
```

include/quic/connection_ids.hpp
jagerman: I suppose it could be `oxen::quic::connection_interface::ReferenceID` but that brings up memory of Java.

include/quic/connection_ids.hpp
```diff
@@ -0,0 +1,108 @@
...
+        ReferenceID(const ReferenceID& obj) = default;
+        ReferenceID(ReferenceID&& obj) = default;
+
+        ReferenceID& operator=(const ReferenceID& obj) = default;
+        ReferenceID& operator=(ReferenceID&& obj) = default;
+
+        inline bool operator<(const ReferenceID& other) const { return id < other.id; }
+
+        explicit operator const uint64_t&() const { return id; }
+        // operator uint64_t&() { return id; }
```
jagerman: Leftover

include/quic/connection_ids.hpp
```diff
@@ -0,0 +1,108 @@
...
+    // connection to be reached at any of n number of connection ID's (see above
+    // constexpr) at a time. As a result, we key each of those QUIC CID's to the
+    // unique reference ID, to which the connection shared_ptr is keyed in the
+    // master map container
+    struct alignas(uint64_t) ReferenceID final
+    {
+        uint64_t id;
+
+        ReferenceID() = delete;
+        ReferenceID(uint64_t v) : id{v} {}
```
jagerman:
```suggestion
        explicit ReferenceID(uint64_t v) : id{v} {}
```

include/quic/connection_ids.hpp
jagerman: From discussion:
- rename `ConnectionID` to `quic_cid` (this is an internal libquic type)
- rename `ReferenceID` to `ConnectionID` for (mostly) external use

jagerman, line 21:
```suggestion
    {
        using value_type = uint64_t;
```

(goes with the std::hash suggestion I'm making)

include/quic/connection_ids.hpp
jagerman: Scratch it, not necessary.

include/quic/connection_ids.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    template <>
+    struct hash<oxen::quic::ReferenceID>
+    {
+        size_t operator()(const oxen::quic::ReferenceID& rid) const
+        {
+            static_assert(
+                    alignof(oxen::quic::ReferenceID) >= alignof(uint64_t) &&
+                    offsetof(oxen::quic::ReferenceID, id) % sizeof(uint64_t) == 0);
+            return *reinterpret_cast<const size_t*>(rid.id);
```
jagerman: Since the reference ID is just a wrapped integer value there's no need for the reinterpret_cast on the data (which is only really something you want for random data -- which is fine for ConnectionID, but not great for a monotonically increasing value).  Instead you can just defer this one to the built-in integer hasher:
```suggestion
            return std::hash<decltype(rid.id)>{}(rid.id);
```

jagerman (approved):

PR #79, opened 2024-01-09
BTRequestStream: generic handler & exception catching

jagerman:
This adds the ability to construct a BTRequestStream with a generic command handler so that if you have a non-trivial set of commands you don't have to incur a potentially large instantiation cost of constructing a bunch of lambda for `register_command` calls that may never be used.  Instead you can now provide a single handler that gets called for any incoming command.

This also adds the ability to throw a new `no_such_endpoint` exception so that such a handler can signal that an endpoint doesn't exist.

Since this necessitated adding an exception handler anyway, this also catches generic exceptions and wraps them with a generic error-producing response function.  Previously an uncaught exception throw from a handler would crash the program.

### DIFF
--- a/include/quic/btstream.hpp
+++ b/include/quic/btstream.hpp
@@ -245,6 +245,12 @@ namespace oxen::quic
         /// commands that are not created via `register_command` calls.
         void register_command(std::string endpoint, std::function<void(message)>);
 
+        /// Registered (or replaces) the fallback command handler that is invoked if the requested
+        /// endpoint does not match any endpoint set up with `register_command`.  If no individual
+        /// `register_command` endpoints are set up at all then this becomes the single callback to
+        /// invoke for all incoming commands.
+        void register_command_fallback(std::function<void(message)> request_handler);
+
         const Address& local() const { return conn.local(); }
 
         const Address& remote() const { return conn.remote(); }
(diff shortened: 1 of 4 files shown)


include/quic/btstream.hpp
```diff
@@ -224,19 +239,33 @@ namespace oxen::quic
...
         void handle_bp_opt(std::function<void(Stream&, uint64_t)> close_cb)
         {
             log::debug(bp_cat, "Bparser set user-provided close callback!");
             close_callback = std::move(close_cb);
         }
 
+        // Optional constructor argument: generic request handler.  If set, this is invoked for all
+        // incoming requests to endpoints that do not have a `register_command`.  (And so if
+        // `register_command` is not used at all, this can handle all endpoint).
+        void handle_bp_opt(std::function<void(oxen::quic::message m)> request_handler)
```
dr7ana: Do you think it would be better to use a member function `register_generic_handler`, similar to how we use `register_command` for each specific endpoint?

I don't feel too strongly either way, but the explicit-ness of the method feels like a convincing idea to me. 

If you didn't want to create a new method, we could even change the parameters of  `register_command` like:
```c++
void BTRequestStream::register_command(std::optional<std::string> ep, std::function<void(message)> func)
{
    endpoint.call(
        [this, ep = std::move(ep), func = std::move(func)]() mutable { 
            if (ep)
                func_map[std::move(ep)] = std::move(func);
            else
                generic_handler = std::move(func);
        });
}
```
So setting the generic handler means passing `std::nullopt` to the first argument.

tests/002-send-receive.cpp
```diff
@@ -628,4 +628,64 @@ namespace oxen::quic::test
...
+    TEST_CASE("002 - BParser generic request handler", "[002][bparser][generic]")
+    {
+        Network test_net{};
+
+        auto [client_tls, server_tls] = defaults::tls_creds_from_ed_keys();
+
+        Address server_local{};
+        Address client_local{};
+
+        std::mutex mut;
```
dr7ana: Unused and triggered a fail in CI [here](https://ci.oxen.rocks/oxen-io/oxen-libquic/503/5/3)

include/quic/btstream.hpp
jagerman: I suppose it could have a method for it as well, but it is pretty simple to be able to just slap it into the constructor.

jagerman: There's a second reason for it being useful in the constructor: we don't have to worry about thread synchronization during construction and so can just set it directly, where the method version (which I'm also adding) has to worry about this by deferring the assignment to an event loop call.

dr7ana (approved):

PR #85, opened 2024-01-22
Quiet connection closure

dr7ana:


tests/001-handshake.cpp
```diff
@@ -283,4 +283,122 @@ namespace oxen::quic::test
...
+
+            // The pubkeys definitely should not be the same
+            REQUIRE(rv != 0);
+
+            // If the LHS parameter to std::strcmp appears FIRST in lexicographical order, then rv < 0. As a result,
+            // if the incoming pubkey appears BEFORE the server pubkey in lexicographical order, we will defer to the
+            // connection initiated by the remote -- ergo we will ACCEPT this connection and mark the local endpoint's
+            // connection as "die silently" (close w/o executing any callbacks or writing any close packets). Else, we
+            // will REJECT the incoming connection and defer to the local endpoint's outgoing connection
+            auto defer_to_incoming = rv < 0;
```
jagerman:
```suggestion
            // The pubkeys definitely should not be the same
            REQUIRE(incoming != local);

            // If the incoming pubkey appears BEFORE the server pubkey in lexicographical order, we will defer to the
            // connection initiated by the remote -- ergo we will ACCEPT this connection and mark the local endpoint's
            // connection as "die silently" (close w/o executing any callbacks or writing any close packets). Else, we
            // will REJECT the incoming connection and defer to the local endpoint's outgoing connection
            auto defer_to_incoming = incoming < local;
```
No need to go to C functions for lexicographic comparison.

jagerman: (The string can also contain `\x00` characters, which the strcmp won't deal with correctly but std::string `<` will)

include/oxen/quic/connection_ids.hpp
```diff
@@ -31,6 +31,9 @@ namespace oxen::quic
 
         inline bool operator<(const ConnectionID& other) const { return id < other.id; }
 
+        inline bool operator==(const ConnectionID& other) const { return id == other.id; }
+        inline bool operator!=(const ConnectionID other) const { return !(*this == other); }
```
jagerman:
```suggestion
        inline bool operator!=(const ConnectionID& other) const { return !(*this == other); }
```

jagerman (approved):

PR #87, opened 2024-01-24
Move IOChannel into its own header

jagerman:
It felt rather strange for it to be embedded in one of its subclass's headers, which meant stream.hpp had to include datagram.hpp just to get at the common base class it shares with datagram.hpp.


dr7ana (approved):
good change :+1:

PR #88, opened 2024-01-26
Stream/connection thread-safety and lifetime fixes

jagerman:
Fixes various issues in Stream, mostly around thread-safety and lifetime.

This all started in Python bindings when a Stream object could easily outlive its Connection (since we really don't have nicely-defined lifetime control there, because destruction depends on Python garbage collection), which could then segfault if you tried to (for instance) send down it.

This turns out to be not just a Python issue, though, as the same problem could occur (if not exactly easily) if you, for instance, held a Stream in a lambda to send something down it.

This snowballed into fixes for various Stream/Connection values being non-threadsafe.

The first few commits here were separable; see their individual descriptions.

High-level summary of the changes in the Big Commit:
- Various thread-unsafe methods are now made protected, with the old public name now becoming a wrapper than safely accesses (or sets) the appropriate value through a `call_get`.
- Stream and datagram_interface now just have a copy of their Connection's reference_id (so that we don't have to synchronously go to the Connection to get it safely).  This also means it can still be available even if the Connection has gone away.
- max datagram size handling got revamped: we have little choice but to drop (rather than throw an exception when trying to send something too big) because of the necessity of checking the maximum inside the `call`, so this necessitated some changes:
  - we now track the last max datagram size (internally) and flip an atomic<bool> if that max size changes.
  - callers can query the last size, or can check it only if it has changed.  (These aren't necessary, but provided in case an application needs to worry about the specific max datagram size).
- Streams no longer hold a `Connection&`, but instead a `Connection*` that can be nullptr'ed by the owning Connection when the Connection is going away.  Stream's now-call_get wrapped methods properly check that they have a Connection before trying to use it, and failing (or dropping) if the connection has gone away.
- Some paths through Connection shutdown didn't `halt_events()`, and some paths could end up trying to packet_io_ready() after a half_events() and segfault (because the packet_io_trigger was already deleted).  Fixed.
- Streams didn't always have their stream close callback fired.  This fixes it by calling a new `drop_streams()` method during connection close firing.  (Similarly to connection close callbacks, we don't fire them if the connection has been set to close quietly).

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -78,7 +78,7 @@ local generic_build(jobs, build_type, lto, werror, cmake_extra, local_mirror, te
         + (if tests then [
              'cd build',
              '../utils/gen-certs.sh',
-             (if gdb then '../utils/ci/drone-gdb.sh ' else '') + './tests/alltests --log-level debug --no-ipv6 --colour-mode ansi',
+             (if gdb then '../utils/ci/drone-gdb.sh ' else '') + './tests/alltests --success --log-level debug --no-ipv6 --colour-mode ansi',
              'cd ..',
            ] else []);
 
--- a/include/oxen/quic/btstream.hpp
+++ b/include/oxen/quic/btstream.hpp
@@ -235,12 +235,6 @@ namespace oxen::quic
 
         void respond(int64_t rid, bstring_view body, bool error = false);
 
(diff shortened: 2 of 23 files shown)


dr7ana (approved):
looks good to me! flipped testnet to this and running smoothly :+1:

PR #90, opened 2024-01-30
Add a compile-time failure if not providing TLS creds to connect(...)

jagerman:
I made a mistake and forgot to pass TLS credentials to `connect()`, but that can never work so this makes it into a compile-time error:

```
/home/jagerman/src/oxen-libquic/include/oxen/quic/endpoint.hpp:114:95: error: static assertion failed: No TLSCreds provided to connect()
  114 |                     (std::is_convertible_v<remove_cvref_t<Opt>, std::shared_ptr<TLSCreds>> || ...),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

### DIFF
--- a/include/oxen/quic/endpoint.hpp
+++ b/include/oxen/quic/endpoint.hpp
@@ -34,6 +34,14 @@ extern "C"
 
 namespace oxen::quic
 {
+    template <typename... Opt>
+    static constexpr void check_for_tls_creds()
+    {
+        static_assert(
+                (0 + ... + std::is_convertible_v<remove_cvref_t<Opt>, std::shared_ptr<TLSCreds>>) == 1,
+                "Endpoint listen/connect require exactly one std::shared_ptr<TLSCreds> argument");
+    }
+
     class Endpoint : public std::enable_shared_from_this<Endpoint>
     {
         friend class TestHelper;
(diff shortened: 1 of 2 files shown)


dr7ana (commented):
Two things:
1) We should do the same with `::listen(...)` in regard to passing a `TLSCreds` object. That would sort-of invalidate [this](https://github.com/jagerman/oxen-libquic/blob/acfa036ff8463618f5b13b286c8415621f4697a1/tests/001-handshake.cpp#L118) test case, as it would then be a compile fail rather than a throw/no-throw thing

2) Do you think it's worth adding an explanatory no-op test case like [this](https://github.com/jagerman/oxen-libquic/blob/acfa036ff8463618f5b13b286c8415621f4697a1/tests/001-handshake.cpp#L159)?

jagerman:
> We should do the same with ::listen(...) in regard to passing a TLSCreds object. That would sort-of invalidate [this](https://github.com/jagerman/oxen-libquic/blob/acfa036ff8463618f5b13b286c8415621f4697a1/tests/001-handshake.cpp#L118) test case, as it would then be a compile fail rather than a throw/no-throw thing

`listen(...)` already has one.  Actually, it also makes sure you only pass 1 instance as well, so maybe I should update the `connect` one to match it.

jagerman:
Updated.

tewinget (approved):

PR #96, opened 2024-01-31
Allow individual handshake timeouts on connections

jagerman:
This adds support for passing `opt::handshake_timeout` into listen/connect so that you can change the timeout for that connection to something different than the endpoint; the endpoint's value now simply becomes the default for connections that don't override it (similarly to how endpoint vs connection connection close callbacks work).

It also maps 0 (or negative) timeouts to ngtcp2's special "no timeout" value of UINT64_MAX.

I also increased the default handshake timeout to 10s (up from 5s) as it seems possible to me that some feasible, if rare, network latency could result in a legitimate 5s handshake time (e.g. think quic over lokinet over a weak mobile internet connection).

### DIFF
--- a/include/oxen/quic/connection.hpp
+++ b/include/oxen/quic/connection.hpp
@@ -262,6 +262,8 @@ namespace oxen::quic
         //          will select the first in the client's list it also supports, so the user
         //          should list them in decreasing priority. If the user does not specify alpns,
         //          the default will be set
+        //      default_handshake_timeout: the default timeout for handshaking for the endpoint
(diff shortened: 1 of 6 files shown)


jagerman:
Some context on why I want this: for storage server reachability tests I want to be able to use a tighter handshake timeout when doing outgoing reachability testing (e.g. 2s) while still allowing longer incoming handshake timeouts (10s) for clients connecting to a storage server, but with only one setting on endpoint I couldn't do that.

PR #100, opened 2024-02-01
Various last minute fixes

jagerman:


tewinget (approved):

jagerman:
[Allow providing an Endpoint static secret](https://github.com/oxen-io/oxen-libquic/pull/100/commits/a8986bfb1b4a01d73db276ace1c05dc99f02d68f) is so that we can use persistent secret data for generating tokens so that packets remain valid across server restarts and we don't have to force clients to retry if the server restarts.

https://github.com/jagerman/loki-storage-server/commit/4cdaf5a27dc210872be8c0b712081234d445eada for the implementation in storage server (which solves the various warnings caused by restarts generating new secret data that then fails the existing initial packet token data).

src/connection.cpp
```diff
@@ -41,272 +41,270 @@ namespace oxen::quic
         }
     }
 
-    int hook_func(
-            gnutls_session_t session, unsigned int htype, unsigned when, unsigned int incoming, const gnutls_datum_t* msg)
+    // ngtcp2 internal callback functions (that are deliberately source only, not a published
+    // header); we group them all in this struct to make them slightly easier to manage, but more
+    // importantly, because `Callbacks` is a friend-with-benefits of Endpoint that can touch its
+    // privates.
+    struct Callbacks
```
dr7ana, line 48: I actually really like this idea for storing the callbacks we use and giving them access. I wonder if it would be useful in the future to have TestHelper extend that class so we could override and test specific callbacks, store other state, etc

src/endpoint.cpp
```diff
@@ -318,8 +322,8 @@ namespace oxen::quic
                         "Error: failed to send close packet [{}]; removing connection ({})",
                         rv.str_error(),
                         conn.reference_id());
+                delete_connection(conn);
```
dr7ana, line 336: Good catch

src/connection.cpp
```diff
@@ -215,14 +215,15 @@ namespace oxen::quic
...
-                        token, conn->static_secret(), NGTCP2_STATELESS_RESET_TOKENLEN, cid) != 0)
+                        token, ep._static_secret.data(), ep._static_secret.size(), cid) != 0)
                 return NGTCP2_ERR_CALLBACK_FAILURE;
 
             auto dir_str = conn->is_outbound() ? "CLIENT"s : "SERVER"s;
             log::trace(log_cat, "{} generated new CID for {}", dir_str, conn->reference_id());
-            conn->endpoint().associate_cid(cid, *conn);
+            ep.associate_cid(cid, *conn);
 
             // TODO: send new stateless reset token
```
dr7ana, line 228: I think this entire commit is a great change, but I'm slightly surprised you want to include it before any of the logic to handle the stateless reset tokens is implemented?  It's pretty easy, a few functions listed in the TODO's, so if you were planning on looping that in it would be totally feasible

dr7ana (approved):

PR #105, opened 2024-02-05
Remove fmt and oxen logging from headers

jagerman:
Having them in headers breaks packaging because it means we would have to build packages for both oxen-logging, spdlog, and fmt, which is a nuissance.

This PR changes the code so that we only use fmt/spdlog/oxen-logging internal (in src/ and test/) but *don't* use or require it from any headers.

There is a tiny bit of fmt-dependent code left in quic/format.hpp, but this is now not included by default and is an optional include an application with fmt available can include to be able to format quic types.

For the most part this just moves logs into the cpp and changes a few formatted strings to work via stl functions, but a few trace logs got dropped because they seemed unimportant enough to move into the cpp file.

### DIFF
--- a/include/oxen/quic.hpp
+++ b/include/oxen/quic.hpp
@@ -7,8 +7,10 @@
 #include "quic/crypto.hpp"
 #include "quic/datagram.hpp"
 #include "quic/endpoint.hpp"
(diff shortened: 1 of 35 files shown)


jagerman:
(This is already live in the 1.0.0 debs, via a patch; merging now to include it in a 1.0.1 release)

PR #106, opened 2024-02-05
Make sure sent_requests have a keep-alive

jagerman:
A close at just the wrong time could result in an invalid string_view still pending on a Stream.

This seems too problematic to have the "no ownership" parameter as a default argument, so I've made it explicit (and added nullptr to the many test suites places previously using it implicitly).

### DIFF
--- a/include/oxen/quic/iochannel.hpp
+++ b/include/oxen/quic/iochannel.hpp
@@ -45,12 +45,17 @@ namespace oxen::quic
         Address local() const;
         Address remote() const;
 
+        // Sends a string_view of character/byte data.  The keep_alive argument is used to manage
+        // ownership: it will be held until the data is completely sent and acked by the remote
+        // side.  It can be nullptr to disable, but only if the caller is certain that the provided
+        // data will stay alive for the duration of the Stream's lifetime.
         template <typename CharType, std::enable_if_t<sizeof(CharType) == 1, int> = 0>
-        void send(std::basic_string_view<CharType> data, std::shared_ptr<void> keep_alive = nullptr)
+        void send(std::basic_string_view<CharType> data, std::shared_ptr<void> keep_alive)
         {
             send_impl(convert_sv<std::byte>(data), std::move(keep_alive));
         }
 
+        // Takes over ownership of a string of character/byte data and sends it into the stream.
         template <typename CharType>
         void send(std::basic_string<CharType>&& data)
         {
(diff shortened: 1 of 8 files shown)


jagerman:
No test case because I haven't a clue how I could trigger this manually.

I'm also not 100% sure that this *is* the issue we saw in Lokinet, but it seems to me that it could be, particularly when combined with the Lokinet close-one-direction logic + RC gossiping.

include/oxen/quic/iochannel.hpp
```diff
@@ -45,12 +45,17 @@ namespace oxen::quic
         Address local() const;
         Address remote() const;
 
+        // Sends a string_view of character/byte data.  The keep_alive argument is used to manage
+        // ownership: it will be held until the data is completely sent and acked by the remote
+        // side.  It can be nullptr to disable, but only if the caller is certain that the provided
+        // data will stay alive for the duration of the Stream's lifetime.
         template <typename CharType, std::enable_if_t<sizeof(CharType) == 1, int> = 0>
-        void send(std::basic_string_view<CharType> data, std::shared_ptr<void> keep_alive = nullptr)
+        void send(std::basic_string_view<CharType> data, std::shared_ptr<void> keep_alive)
```
dr7ana, line 53: I think this is a good change, thanks for knocking this out yesterday! I have one reservation that's slightly more than a nit-pick in regards to defaulting the shared_ptr to nullptr.

I understand the desire to explicitly require the caller to ensure the data will stay alive, but wasn't the whole point of introducing that parameter to do that implicitly? I just wonder if there's a way we can do what you're implementing without having to explicitly write `nullptr` on every call to `Stream::send`.

If the issue was with string_views, why couldn't we do something like this (similar to the overloads of `::send` that take a string-type and vector):
```C++
template <typename CharType, std::enable_if_t<sizeof(CharType) == 1, int> = 0>
void send(std::basic_string_view<CharType> data, std::shared_ptr<void> keep_alive = nullptr)
{
    std::string string_data{data};
    if (not keep_alive)
        keep_alive = std::make_shared<std::basic_string<CharType>>(std::move(string_data));
    send(std::string_view{*string_data}, std::move(keep_alive));
}
```
(Apologies if this isn't exactly targeted at the purpose behind this PR, I believe I missed most of that discussion between you and Tom)

jagerman: > I understand the desire to explicitly require the caller to ensure the data will stay alive, but wasn't the whole point of introducing that parameter to do that implicitly? I just wonder if there's a way we can do what you're implementing without having to explicitly write nullptr on every call to Stream::send.

If you give it a string (with no nullptr) then that's exactly what you get.  This is only for the version of send taking a `std::string_view`.

jagerman: > If the issue was with string_views, why couldn't we do something like this:

There are cases where we don't need/want a keep-alive (e.g. because we have some application-side code that guarantees it'll stay alive), and that's what the nullptr allows.  The test suite isn't a *great* example of this because it's full of long-lived static strings, but other applications can have that issue too.

PR #107, opened 2024-02-20
Downgrade warning firing on connection timeouts

jagerman:
There are way too many of these in storage serever logs in practice just from connection testing to endpoints that aren't currently reachable, and this isn't (necessarily) indicative of a local problem.  This downgrades it to debug; if the application is doing something here that should be warned about then it can warn about it in a close handler.


jagerman:
(Already in live 1.0.0 debs; merging for inclusion in 1.0.1)

PR #109, opened 2024-03-20
Incorrect pubkey length crash

dr7ana:
- When initiating an outbound connection, passing an ed25519 pubkey of incorrect length will result in an exception being thrown and the endpoint shutting down/closing all connections, as designed
- As not designed, the endpoint will see the ConnectionID mapped to a nullptr (as the connection object failed to be created) and will segfault on bad memory access
- 001 test case added for incorrect keylenths
- Thanks @mpretty-cyro for the catch!

### DIFF
--- a/include/oxen/quic/endpoint.hpp
+++ b/include/oxen/quic/endpoint.hpp
@@ -97,19 +97,22 @@ namespace oxen::quic
             Path _path = Path{_local, remote};
 
             net.call([&opts..., &p, path = _path, this, remote_pk = std::move(remote).get_remote_key()]() mutable {
+                quic_cid qcid;
+                auto next_rid = next_reference_id();
+
                 try
                 {
                     // initialize client context and client tls context simultaneously
                     outbound_ctx = std::make_shared<IOContext>(Direction::OUTBOUND, std::forward<Opt>(opts)...);
                     _set_context_globals(outbound_ctx);
 
-                    auto next_rid = next_reference_id();
-
                     for (;;)
                     {
                         // emplace random CID into lookup keyed to unique reference ID
                         if (auto [it_a, res_a] = conn_lookup.emplace(quic_cid::random(), next_rid); res_a)
                         {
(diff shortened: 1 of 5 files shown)


include/oxen/quic/endpoint.hpp
```diff
@@ -97,19 +97,22 @@ namespace oxen::quic
...
                     _set_context_globals(outbound_ctx);
 
-                    auto next_rid = next_reference_id();
-
                     for (;;)
                     {
                         // emplace random CID into lookup keyed to unique reference ID
                         if (auto [it_a, res_a] = conn_lookup.emplace(quic_cid::random(), next_rid); res_a)
                         {
+                            qcid = it_a->first;
```
tewinget, line 114: If something before here can `throw`, the `erase` in the `catch` below will be using `qcid` uninitialized.  If not, or if not in a way we want/care to `catch` here, we could move the `try` to just before the `for`.  Alternatively (and perhaps regardless), `qcid` should be initialized above to its zero-equivalent.

dr7ana: I believe it's already calling the default constructor -- `quic-cid` has one while `ConnectionID` explicitly deletes its constructor.  You don't need the braces to implicitly call the default constructor (I think that's a C-ism). I checked in the examples here and it looks correct?
https://en.cppreference.com/w/cpp/language/default_initialization
https://en.cppreference.com/w/cpp/language/default_constructor

tewinget: Yeah, looking it up it does seem that way.  It seems I've been mislead seeing so many objects declared with empty `{}` after the var name.  I'm maybe mis-remembering doing this for specifically primitive types.  Ah well.

dr7ana: Worth mentioning to do this exploration into cppreference :+1:

tewinget (approved):

PR #111, opened 2024-04-02
Update CMake files to support iOS builds

mpretty-cyro:


cmake/StaticBuild.cmake
```diff
@@ -259,38 +271,50 @@ if(APPLE AND CMAKE_CROSSCOMPILING)
...
-elseif(gmp_build_host STREQUAL "")
+elseif(gmp_build_host STREQUAL "" AND CMAKE_LIBRARY_ARCHITECTURE)
     set(gmp_build_host "--build=${CMAKE_LIBRARY_ARCHITECTURE}")
 endif()
 
-
-
 build_external(libtasn1
-    CONFIGURE_EXTRA --disable-doc
+    CONFIGURE_COMMAND ./configure ${gmp_build_host} --disable-shared --disable-doc --prefix=${DEPS_DESTDIR} --with-pic
```
jagerman: We should probably rename `gmp_build_host` since it's now being used for other packages as well.

mpretty-cyro: I've renamed it to `build_host` instead

PR #113, opened 2024-04-12
Windows fixes (and various other bits)

jagerman:
This fixes various small issues on Windows to get the test suite working (via wine).

New useful things added here:
- The `alltests` binary now has a `-T` flag that can print a trace of all test case and sections being entered.
- The windows build now runs the test suite under wine.

Fixes:
- If a socket option failed, the program could crash because we weren't checking for Connection shutdown in between socket operations and rescheduling a transmit timer (if such a socket operation failed, it would trigger Connection shutdown, invalidating the retransmit timer pointer).
- `event_get_supported_methods` is not thread safe; we now call it just once (via a static local variable) instead of during each Loop construction to avoid such crash potential.
- Windows binaries were not fully static (they required libstdc++-6.dll and similar, which we don't easily have or want to ship alongside the binaries)
- Dual stack IPv6 was not working on Windows (or anywhere else outside Linux/macOS).  This commit now does it properly so that it is either explicitly enabled or disabled depending on the Address value that we bind.
- IPv4 address tests were broken (and ifdef'd away) on Windows; this includes a Windows-specific hack to get it working.
- WSACleanup() was called during `Loop` destruction, but *also* during `Loop::shutdown`, but WSACleanup calls must be one-to-one with `WSAStartup` calls.  This broken the test suite on Windows because in cases where we have two Network objects, the first one being destroyed could WSACleanup *twice*, which then completely stalled the second one's event loop which would then deadlock.  This fixes it to always unconditionally call it (just once) in the destructor, matching the WSAStartup call in the constructor.
- socket operation failures did not have a log statement: they now have a descriptive error with what failed, the error code, and the error reason.
- Windows IPv4 bound address were failing because Windows defines a `IP_RECVDSTADDR` symbol that it doesn't actually support.  This fixes it to ignore `IP_RECVDSTADDR` on Windows (we need to use `IP_PKTINFO` instead).
- Fixed `ipv4::operator in_addr` compatibility in Windows, where in_addr is non-simple enough that the reinterpret_cast was failing strict aliasing rules.

Random cleanups:
- added a format rule to eliminate unnecessary semicolons
- cleaned up the C++20 formattable concept using a macro instead of an inline ifdef for the `bool` that older GCC needs.  (This is the same macro name that is now being used in Lokinet, which can just drop its own definition once it updates to this oxen-libquic).
- Changed the default speedtest-server/-client log level to info instead of debug.  Debug is pointless as a default because it just bottlenecks the CPU printing log statements when sending huge numbers of packets.
- Fixed mismatched order of declared variables on Windows and non-Windows code paths

### DIFF
--- a/include/oxen/quic/address.hpp
+++ b/include/oxen/quic/address.hpp
@@ -4,6 +4,19 @@
 #include "ip.hpp"
 #include "utils.hpp"
 
+#if defined(__OpenBSD__) || defined(__DragonFly__)
+// These systems are known to disallow dual stack binding, and so on such systems when
+// invoked with an empty address we default to the IPv4 any address rather than the IPv6 any
+// address as the IPv4 is very likely to be more usable.
+//
+// A libquic-using application that wants to support full dual stack on these OSes as well
+// will need to modify how they use QUIC to maintain *two* endpoints: one bound to `[::]`
+// and one bound to `0.0.0.0`.  (Using such an explicit address instead of a
+// default-constructed or empty IP address will not be dual stack anywhere).  Alternatively an
+// application can use this OXEN_LIBQUIC_ADDRESS_NO_DUAL_STACK definition to figure something out.
+#define OXEN_LIBQUIC_ADDRESS_NO_DUAL_STACK
+#endif
+
 namespace oxen::quic
 {
     inline constexpr std::array<uint8_t, 16> _ipv6_any_addr = {0};
@@ -22,10 +35,14 @@ namespace oxen::quic
         {
             std::memmove(&_sock_addr, &obj._sock_addr, sizeof(_sock_addr));
             _addr.addrlen = obj._addr.addrlen;
+            dual_stack = obj.dual_stack;
         }
 
       public:
-        // Default constructor or single-port constructor yields [::]:port (or [::]:0 if port omitted)
+        /// Default constructor or single-port constructor yields [::]:port (or [::]:0 if port
+        /// omitted) on most platforms where dual-stack IPv6/IPv4 sockets are supported.  On OSes
+        /// that do not allow dual-stack sockets (OpenBSD, DragonFlyBSD) the default for an empty IP
+        /// address is instead the IPv4 any address (0.0.0.0).
         explicit Address(uint16_t port = 0) : Address{"", port} {}
 
         Address(const sockaddr* s, socklen_t n)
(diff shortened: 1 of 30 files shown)


tewinget (approved):

include/oxen/quic/address.hpp
```diff
@@ -4,6 +4,19 @@
...
+// These systems are known to disallow dual stack binding, and so on such systems when
+// invoked with an empty address we default to the IPv4 any address rather than the IPv6 any
+// address as the IPv4 is very likely to be more usable.
+//
+// A libquic-using application that wants to support full dual stack on these OSes as well
+// will need to modify how they use QUIC to maintain *two* endpoints: one bound to `[::]`
+// and one bound to `0.0.0.0`.  (Using such an explicit address instead of a
+// default-constructed or empty IP address will not be dual stack anywhere).  Alternatively an
+// application can use this OXEN_QUIC_ADDRESS_NO_DUAL_STACK definition to figure something out.
+#define OXEN_QUIC_ADDRESS_NO_DUAL_STACK
```
dr7ana: Super minor: The other compile options we have in libquic start with `LIBQUIC_{...}`

Thoughts on making this `LIBQUIC_ADDRESS_NO_DUAL_STACK` for continuity?

include/oxen/quic/ip.hpp
```diff
@@ -21,9 +21,12 @@ namespace oxen::quic
 
         const std::string to_string() const;
 
-        explicit operator const in_addr&() const { return reinterpret_cast<const in_addr&>(addr); }
-
-        explicit operator const uint32_t&() const { return reinterpret_cast<const uint32_t&>(addr); }
+        explicit operator in_addr() const
```
dr7ana, line 24: I saw this was in the commit titled `address aliasing warning under minGW`. Could you please elaborate why we need to change this to returning a copy? I wanted this conversion operator to use the `ipv4` object as an `in_addr` without copying

dr7ana (commented):
Thanks for busting this out!

include/oxen/quic/ip.hpp
jagerman: Under mingw, the reinterpret_cast here isn't valid because the `in_addr` type isn't just a struct with a uint32_t in it--it's instead a union of a few different representations.

But also: because `in_addr` itself is such a simple, primitive type (it's just struct with a single uint32_t in it, even under Windows -- it's just unioned with other things under Windows) returning by reference doesn't optimize anything: you're either returning a pointer (for the reference) or a 4-byte primitive.

dr7ana: Makes perfect sense 👍

include/oxen/quic/address.hpp
jagerman: Renamed to OXEN_LIBQUIC_ADDRESS_NO_DUAL_STACK to match the other definitions.

PR #114, opened 2024-04-12
No-UDP Endpoints: manual data I/O routing

dr7ana:
- Now you can provide a callback that bypasses sending packets through the UDP socket
- This also disables creation of the UDP socket entirely
- The application will need to take responsibility for passing packets into the Endpoint via Endpoint::manually_receive_packet(...)

### DIFF
--- a/include/oxen/quic/endpoint.hpp
+++ b/include/oxen/quic/endpoint.hpp
@@ -57,8 +57,8 @@ namespace oxen::quic
         template <typename... Opt>
         Endpoint(Network& n, const Address& listen_addr, Opt&&... opts) : net{n}, _local{listen_addr}
         {
-            _init_internals();
             ((void)handle_ep_opt(std::forward<Opt>(opts)), ...);
+            _init_internals();
             if (_static_secret.empty())
                 _static_secret = make_static_secret();
         }
@@ -88,7 +88,7 @@ namespace oxen::quic
             std::promise<std::shared_ptr<Connection>> p;
             auto f = p.get_future();
 
-            if (!remote.is_addressable())
+            if (not _manual_routing and !remote.is_addressable())
                 throw std::invalid_argument("Address must be addressible to connect");
 
             if (_local.is_ipv6() && !remote.is_ipv6())
@@ -195,6 +195,8 @@ namespace oxen::quic
         // Returns a random value suitable for use as the Endpoint static secret value.
         static ustring make_static_secret();
 
+        void manually_receive_packet(Packet&& pkt);
+
       private:
         friend class Network;
         friend class Loop;
@@ -211,6 +213,8 @@ namespace oxen::quic
         bool _packet_splitting{false};
         Splitting _policy{Splitting::NONE};
         int _rbufsize{4096};
+
+        opt::manual_routing _manual_routing;
 
         uint64_t _next_rid{0};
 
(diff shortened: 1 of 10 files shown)


include/oxen/quic/opt.hpp
```diff
@@ -6,121 +6,157 @@
...
 #include "types.hpp"
 
-namespace oxen::quic::opt
+namespace oxen::quic
 {
-    using namespace std::chrono_literals;
+    class Endpoint;
 
-    struct max_streams
+    namespace opt
```
jagerman, line 13: The extra indentation here is a bit annoying; perhaps this could be:
```C++

namespace oxen::quic { class Endpoint; }
namespace oxen::quic::opt
{
...
}
```
to keep it less annoying.

include/oxen/quic/endpoint.hpp
```diff
@@ -88,7 +88,7 @@ namespace oxen::quic
             std::promise<std::shared_ptr<Connection>> p;
             auto f = p.get_future();
 
-            if (!remote.is_addressable())
+            if (not _manual_routing and !remote.is_addressable())
```
jagerman, line 91:
```suggestion
            if (!_manual_routing && !remote.is_addressable())
```
The mixed keyword and non-keyword operators are jarring (and currently everything else in libquic uses symbols instead of keywords).

Side note: I looked for a way to get clang-format to enforce this, but couldn't find one.

src/udp.cpp
```diff
@@ -237,7 +237,18 @@ namespace oxen::quic
...
+                [](evutil_socket_t, short, void* self) {
+#ifndef NDEBUG
+                    check_rv(
+#endif
+                            static_cast<UDPSocket*>(self)
+                                    ->receive()
+#ifndef NDEBUG
+                                    .error_code,
+                            "udp::receive()")
+#endif
```
jagerman, line 265: I don't think this should be `check_rv`'ed for debug builds: it will throw upon failure, which makes the debug and non-debug paths through here substantially different.

jagerman: Checking the return value and logging is fine, just not via `check_rv` because of that throwing behaviour.

src/udp.cpp
```diff
@@ -508,14 +519,20 @@ namespace oxen::quic
...
             std::memset(&source_addr, 0, sizeof(source_addr));
             if (source_ipv4)
+            {
 #ifdef _WIN32
                 source_addr.v4.ipi_addr
 #else
                 source_addr.v4.ipi_spec_dst
 #endif
                         = path.local.in4().sin_addr;
+                source_addr.v4.ipi_ifindex = 0;
```
jagerman: Not needed: this will already be 0 via the memset.

src/udp.cpp
```diff
@@ -508,14 +519,20 @@ namespace oxen::quic
...
+            }
             else
+            {
                 source_addr.v6.ipi6_addr = path.local.in6().sin6_addr;
+                source_addr.v6.ipi6_ifindex = 0;
```
jagerman: Already 0 via the memset

src/udp.cpp
```diff
@@ -740,8 +756,7 @@ namespace oxen::quic
             next_buf += bufsize[i];
 
             rv = sendmsg(sock_, &hdr, 0);
-            if (rv < 0)
-                break;
+            check_rv(rv, "sendmsg");
```
jagerman: We shouldn't throw here.

jagerman (approved):

PR #116, opened 2024-04-15
Bidir alpns

### LINKED ISSUE #103: Add bidirection alpns setting
Currently there's `opt::inbound_alpns` and `opt::outbound_alpns`, because Lokinet needs to distinguish, but it might be nice to have a `opt::alpns{}` that set both at once (for storage server that just uses one single value).

jagerman:
Fixes #103

Adds `opt::alpns{...}` which is equivalent to specifying both `opt::inbound_alpns{...}` and `outbound_alpns{...}` but is very often what a bidirectional endpoint needs (and in many cases is fine even for a unidirectional endpoint) and thus simplifies what calling code needs to do to specify its ALPN.

This also adds a convenience constructor to all three alpns options that allows passing a single string-like (not `ustring`), to simplify invoking the typical case where there is just a single plain text ALPN value.

Thus a typical ALPNs-providing endpoint invocation can be simplified from:
```C++
    net->endpoint(...,
        opt::inbound_alpns{{"MyALPN"_us}},
        opt::outbound_alpns{{"MyALPN"_us}});
```
to just:
```C++
    net->endpoint(..., opt::alpns{"MyALPN"});
```
but leaves the separate opt::in/outbound_alpns for advanced cases where the ALPNs need to differ based on direction.

There's also a commit in here to drop a couple C++20 backports (starts_width/ends_with and remove_cvref_t) that are no longer needed.  (These seem random, but came along with a change I was making for the other commit here that ended up not being needed -- but are still worthwhile keeping).

### DIFF
--- a/tests/009-alpns.cpp
+++ b/tests/009-alpns.cpp
@@ -113,6 +113,40 @@ namespace oxen::quic::test
             REQUIRE(client_established2.wait());
             REQUIRE(conn2->selected_alpn() == "relay"_usv);
         }
+
+        SECTION("Bidirectional ALPN incoming")
+        {
+            opt::alpns server_alpns{"special-alpn"};
+
+            auto server_endpoint = test_net.endpoint(server_local, server_alpns, timeout);
+            REQUIRE_NOTHROW(server_endpoint->listen(server_tls));
+
+            RemoteAddress client_remote{defaults::SERVER_PUBKEY, "127.0.0.1"s, server_endpoint->local().port()};
+
+            opt::outbound_alpns client_alpns{{"foobar"_us}};
+            auto client_endpoint = test_net.endpoint(client_local, client_established, client_closed, client_alpns, timeout);
+
+            auto conn = client_endpoint->connect(client_remote, client_tls);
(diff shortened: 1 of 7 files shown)


tests/009-alpns.cpp
```diff
@@ -113,6 +113,40 @@ namespace oxen::quic::test
...
+            auto server_endpoint = test_net.endpoint(server_local, server_alpns, timeout);
+            REQUIRE_NOTHROW(server_endpoint->listen(server_tls));
+
+            RemoteAddress client_remote{defaults::SERVER_PUBKEY, "127.0.0.1"s, server_endpoint->local().port()};
+
+            opt::outbound_alpns client_alpns{{"foobar"_us}};
+            auto client_endpoint = test_net.endpoint(client_local, client_established, client_closed, client_alpns, timeout);
+
+            auto conn = client_endpoint->connect(client_remote, client_tls);
+            CHECK(client_closed.wait(2s));
```
tewinget, line 130: should this be waiting on `client_established` (to fail)?

dr7ana (approved):

PR #120, opened 2024-04-17
Allowing specifying a dgram data callback in listen/connect

jagerman:
It felt restrictive (and a little unintuitive) to have to specify the dgram callback at the endpoint because then I can't have a different callback for individual connections or for incoming versus outgoing connections.  It also feels a little weird because a default stream data callback, in contrast, is only at the listen/connect level but not at the endpoint level.

This commit allows also specifying the datagram callback to `connect()` or `listen()` -- if specified as such, it takes precedence over the one given to endpoint.

### DIFF
--- a/include/oxen/quic/context.hpp
+++ b/include/oxen/quic/context.hpp
@@ -42,6 +42,7 @@ namespace oxen::quic
         stream_open_callback stream_open_cb;
         stream_close_callback stream_close_cb;
         stream_constructor_callback stream_construct_cb;
+        dgram_data_callback dgram_data_cb;
         connection_established_callback conn_established_cb;
         connection_closed_callback conn_closed_cb;
         user_config config{};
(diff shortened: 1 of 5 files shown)


dr7ana (commented):
This PR is a great  idea. I think it would be good to make a test case specifically on your last point "if specified as such, it takes precedence over the one given to endpoint."

We could pass a dgram callback to `::endpoint()` and another to `::connect()`, then verify that nothing was received at the endpoint-level one?

jagerman:
> I think it would be good to make a test case specifically on your last point

Added.

dr7ana:
Just thinking out loud here after our conversations about this over the last week:

We still have that weird asymmetry for endpoints that have only called `::listen(...)` vs those that have not. Any callback passed to `::listen(...)` is effectively an endpoint level callback, so we can't really set connection-specific hooks for server endpoints

jagerman:
> Just thinking out loud here after our conversations about this over the last week:
> 
> We still have that weird asymmetry for endpoints that have only called `::listen(...)` vs those that have not. Any callback passed to `::listen(...)` is effectively an endpoint level callback, so we can't really set connection-specific hooks for server endpoints

Agreed, but it's out of the scope of this PR.

dr7ana (approved):

PR #123, opened 2024-04-26
Bump oxen-encoding for constexpr _hex and C++20 changes

jagerman:
Adds more C++20 usage:

- "..."_bsv and "..."_usv are now constexpr
- updating oxen-encoding for constexpr _hex, etc.
- replaces all the `sizeof(Char) == 1` stuff with the new oxenc::basic_char concept (which is more robust)
- replaces `std::enable_if_t`s with `requires` and/or `concepts`
- makes GNUTLSCreds construction take string_views instead of copying std::strings


dr7ana (commented):
Since we don't need `CONCEPT_COMPAT` anymore, would you mind scrapping it with this PR?

dr7ana (approved):

PR #127, opened 2024-05-07
Stream Stop Read/Write

dr7ana:


tests/012-stream_signalling.cpp
```diff
@@ -145,4 +145,66 @@ namespace oxen::quic::test
...
+        REQUIRE(client_stream->is_reading());
+        REQUIRE(client_stream->is_writing());
+
+        SECTION("Stop Writing")
+        {
+            server_stream->stop_writing();
+            REQUIRE_FALSE(server_stream->is_writing());
+
+            client_stream->send(bstring_view{req_msg});
+            REQUIRE(f.get());
```
jagerman: If this fails the test can hang forever; changing this to `require_future(f);` will prevent that.

tests/012-stream_signalling.cpp
```diff
@@ -145,4 +145,66 @@ namespace oxen::quic::test
...
+
+            // allow the acks to get back to the client; extra time for slow CI archs
+            std::this_thread::sleep_for(250ms);
+
```
jagerman, line 211: At this point the client should pass a `REQUIRE_FALSE(client_stream->is_reading())` because the server having shutdown writing should have sent a `STREAM_RESET` to the client, so the client now knows that the read side of the stream is now dead.  (But I don't think this would currently pass).

src/stream.cpp
```diff
@@ -240,28 +301,19 @@ namespace oxen::quic
 
         auto sz = size();
 
+        if (not _is_writing and _unacked_size == 0)
```
jagerman: I think that this should be:
```suggestion
        if (not _is_writing and user_buffers.empty())
```
because I think it's possible to have *no* unacked data, but still have some buffered data that hasn't been sent yet (perhaps because sending is paused, or just because of very unlucky network conditions).  If there is something in user_buffers, however, then that implies that there is something either unsent or unacked (or both).

src/stream.cpp
```diff
@@ -127,11 +127,50 @@ namespace oxen::quic
...
+    void Stream::stop_writing()
+    {
+        endpoint.call([this]() {
+            if (not _is_writing)
+            {
+                log::warning(log_cat, "Stream has already halted write operations!");
+                return;
+            }
+
+            _is_writing = false;
```
jagerman, line 199: This needs to check whether the "can we shutdown write" condition (`user_buffers.empty()` — see other comment) is already satisfied and, if so, initiate the shutdown immediately.  Otherwise, if you shutdown after everything you sent has already been acked, there will never be another ack to trigger the actual stream shutdown code.

jagerman (commented):
Left some comments in the PR.

A couple other things are needed here as well:

1. Attempting to append stream data to a stream that has writing shut down should warn and drop the data, rather than making new buffers for it and storing it.

2. A callback mechanism is needed for an application to know when the stream has entered non-read or non-write state as a result of the other side's actions.  For instance, embedded lokinet's tunneled quic streams are going to need to make the appropriate `shutdown` call in response to the other side shutting down one direction of the stream.

dr7ana:
@jagerman thanks for the feedback, changes made!

PR #130, opened 2024-05-17
Bugfix for vector sender; path methods

dr7ana:
- When passing an r-value reference to ::send_datagram(...), we must make the shared_ptr keep-alive prior to constructing the string_view pointing into the data
- Path comparison methods added plus hash for indexing using paths

### DIFF
--- a/include/oxen/quic/address.hpp
+++ b/include/oxen/quic/address.hpp
@@ -246,6 +246,7 @@ namespace oxen::quic
                     return r <=> 0;
                 return oxenc::big_to_host(a.sin_port) <=> oxenc::big_to_host(b.sin_port);
             }
+
             if (is_ipv6() && other.is_ipv6())
             {
                 auto& a = in6();
@@ -258,7 +259,6 @@ namespace oxen::quic
         }
 
         bool operator==(const Address& other) const { return (*this <=> other) == 0; }
-        bool operator!=(const Address& other) const { return !(*this == other); }
 
         // Returns the size of the sockaddr
         socklen_t socklen() const { return _addr.addrlen; }
@@ -340,6 +340,8 @@ namespace oxen::quic
             return *this;
         }
 
+        bool operator==(const Path& other) const { return std::tie(local, remote) == std::tie(other.local, other.remote); }
+
         // template code to pass Path as ngtcp2_path into ngtcp2 functions
         template <typename T>
             requires std::same_as<T, ngtcp2_path>
@@ -387,6 +389,17 @@ namespace std
 
             auto h = hash<string_view>{}(addr_data);
             h ^= hash<decltype(port)>{}(port) + inverse_golden_ratio + (h << 6) + (h >> 2);
+            return h;
+        }
+    };
+
+    template <>
+    struct hash<oxen::quic::Path>
+    {
+        size_t operator()(const oxen::quic::Path& addr) const
+        {
+            auto h = hash<oxen::quic::Address>{}(addr.local);
+            h ^= hash<oxen::quic::Address>{}(addr.remote) + inverse_golden_ratio + (h << 6) + (h >> 2);
             return h;
         }
     };
--- a/include/oxen/quic/connection.hpp
+++ b/include/oxen/quic/connection.hpp
@@ -129,9 +129,9 @@ namespace oxen::quic
         template <oxenc::basic_char Char>
         void send_datagram(std::vector<Char>&& buf)
         {
-            send_datagram(
-                    std::basic_string_view<Char>{buf.data(), buf.size()},
-                    std::make_shared<std::vector<Char>>(std::move(buf)));
+            auto keep_alive = std::make_shared<std::vector<Char>>(std::move(buf));
+            std::basic_string_view<Char> view{keep_alive->data(), keep_alive->size()};
+            send_datagram(view, std::move(keep_alive));
         }
 
         template <oxenc::basic_char CharType>


include/oxen/quic/address.hpp
```diff
@@ -340,6 +340,10 @@ namespace oxen::quic
             return *this;
         }
 
+        bool operator==(const Path& other) const { return std::tie(local, remote) == std::tie(other.local, other.remote); }
```
tewinget, line 343: Might be worth it/better to do `operator<=>` instead?

dr7ana: great point!

dr7ana: but will define spaceship separately and invoke it within `operator==`, just to follow the Address  pattern

include/oxen/quic/connection.hpp
```diff
@@ -129,9 +129,9 @@ namespace oxen::quic
         template <oxenc::basic_char Char>
         void send_datagram(std::vector<Char>&& buf)
         {
-            send_datagram(
-                    std::basic_string_view<Char>{buf.data(), buf.size()},
-                    std::make_shared<std::vector<Char>>(std::move(buf)));
+            auto keep_alive = std::make_shared<std::vector<Char>>(std::move(buf));
+            std::basic_string_view<Char> view{keep_alive->data(), keep_alive->size()};
+            send_datagram(view, std::move(keep_alive));
```
jagerman, line 134: Though this change won't hurt, it isn't necessary, either.  From https://en.cppreference.com/w/cpp/container/vector/vector: 

> Notes
> 
> After container move construction (overload (8)), references, pointers, and iterators (other than the end iterator) to other remain valid, but refer to elements that are now in *this

i.e. the move transfers ownership, but does not invalidate the pointer to the data.  (This isn't the case for string, however, because of small-string optimization).

include/oxen/quic/address.hpp
```diff
@@ -390,4 +398,15 @@ namespace std
...
         }
     };
+
+    template <>
+    struct hash<oxen::quic::Path>
+    {
+        size_t operator()(const oxen::quic::Path& addr) const
+        {
+            auto h = hash<oxen::quic::Address>{}(addr.local);
+            h ^= hash<oxen::quic::Address>{}(addr.remote);
```
jagerman: See the hash combining just above this using a quasi-random value (`inverse_golden_ratio`) and some bit shifting.  You want something like that when combining hashes so that correlation of the two values generally doesn't result in correlation of the hash.  With just a plain xor you get these undesirable characteristics:
```C++
Address a = /* anything */;
Address b = /* something else */;
hash(Path{a, b}) == hash(Path{b, a})
hash(Path{a, a}) == hash(Path{b, b}) == 0
```
but using the same `h ^= hash<oxen::quic::Address>{}(addr.remote) + inverse_golden_ratio + (h << 6) + (h >> 2);` here avoids such correlations.

include/oxen/quic/address.hpp
```diff
@@ -340,6 +342,12 @@ namespace oxen::quic
             return *this;
         }
 
+        auto operator<=>(const Path& other) const { return std::tie(local, remote) <=> std::tie(other.local, other.remote); }
+
+        bool operator==(const Path& other) const { return (*this <=> other) == 0; }
+
+        bool operator!=(const Path& other) const { return !(*this == other); }
```
jagerman: With C++20, you can drop the `operator !=` and get it automatically via `operator ==`.

include/oxen/quic/address.hpp
```diff
@@ -340,6 +342,12 @@ namespace oxen::quic
             return *this;
         }
 
+        auto operator<=>(const Path& other) const { return std::tie(local, remote) <=> std::tie(other.local, other.remote); }
```
jagerman: `==` (and `!=`) make sense to me, but I'm less convinced that general lexicographic ordering make sense on a Path object.  To elaborate, if I have two paths `p1` and `p2`, then `p1 < p2` doesn't have an obvious meaning.  It's a bit like `std::complex<double>` in that regard: yes, you *could* define inequality operations, but they wouldn't have an obvious meaning.

include/oxen/quic/connection.hpp
dr7ana: This is the exact bug we encountered (the three of us) in the tunnel binaries. I'm just lifting the fix from the WIP PR

include/oxen/quic/address.hpp
dr7ana: It made sense to me when thinking about a local endpoint using paths as key that lead to different remotes:
```c++
Address local_addr{...};

Path path1{local_addr, {remote_addr1, 5555};
Path path2{local_addr, {remote_addr2, 5555}};
Path path3{local_addr, {remote_addr2, 7777}};
```

Using `std::tie` will compare `local` vs `other.local`, then `remote` vs `other.remote` using the `operator<=>` already defined for the `Address` object.

This would result in:

```c++
(path2 > path1) == true;
(path3 > path1) != true;
(path3 > path2) == true;
```

What do you think?

include/oxen/quic/address.hpp
dr7ana: Great! Will drop the one on `Address` too, was just going for uniformity :+1:

include/oxen/quic/address.hpp
jagerman: A local and remote is essentially a 2-dimensional object: you can think of `local` as the `x` dimension and `remote` as the `y` dimension, but an inequality operator between two such instances is a one-dimensional concept that you're applying to two-dimensional space.  You *can* define an order in such a space (such as lexicographic ordering of x-then-y), but it's an arbitrary definition rather than a natural one.

In particular, while x-then-y (i.e. local-then-remote) is indeed *a* valid ordering, it isn't intuitive to me that it is an ordering that is usually preferable to all other orderings.  Lexicographic y-then-x ordering, for instance, seems like another worthy contender, but there are other possibilities (for instance: putting paths with all-private addresses before ones with public components).  When you have more than one different, but reasonable, orderings I don't think it makes sense to define order at all; code that wants to order instances of the objects is then responsible for figuring out the ordering that makes sense in its context.

jagerman (approved):

PR #131, opened 2024-05-17
Max IP in range; Revamp ipv4/6 methods

dr7ana:
- iprange now has ::max_ip() method
- ipv4 enforces internal host-order storage
- methods in other objects now understand this
- [001] updated

### DIFF
--- a/include/oxen/quic/ip.hpp
+++ b/include/oxen/quic/ip.hpp
@@ -7,14 +7,15 @@
 
 namespace oxen::quic
 {
+    inline constexpr size_t IPV4_ARRLEN{4};
+    inline constexpr size_t IPV6_ARRLEN{16};
+
     struct ipv4
     {
-        // Stored in network order
+        // host order
         uint32_t addr;
 
         ipv4() = default;
-        // Constructor using host order IP address string
-        ipv4(const std::string& ip);
 
         constexpr ipv4(uint32_t a) : addr{a} {}
         constexpr ipv4(uint8_t a, uint8_t b, uint8_t c, uint8_t d) :
@@ -26,17 +27,15 @@ namespace oxen::quic
         explicit operator in_addr() const
         {
             in_addr a;
-            a.s_addr = addr;
+            a.s_addr = oxenc::load_host_to_big<uint32_t>(&addr);
             return a;
         }
 
-        constexpr bool operator==(const ipv4& a) const { return addr == a.addr; }
+        constexpr auto operator<=>(const ipv4& a) const { return addr <=> a.addr; }
 
-        constexpr bool operator==(const in_addr& a) const { return addr == a.s_addr; }
+        constexpr bool operator==(const ipv4& a) const { return (addr <=> a.addr) == 0; }
 
-        constexpr bool operator!=(const ipv4& a) const { return not(*this == a); }
-
-        constexpr bool operator<(const ipv4& a) const { return addr < a.addr; }
+        constexpr bool operator==(const in_addr& a) const { return (addr <=> a.s_addr) == 0; }
 
         constexpr ipv4 to_base(uint8_t mask) const { return mask < 32 ? ipv4{(addr >> (32 - mask)) << (32 - mask)} : *this; }
     };
@@ -46,13 +45,15 @@ namespace oxen::quic
         ipv4 base;
         uint8_t mask;
 
+        ipv4 max_ip();
+
+        constexpr ipv4_net(ipv4 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}
+
         const std::string to_string() const { return base.to_base(mask).to_string(); }
 
-        constexpr bool operator==(const ipv4_net& a) const { return base == a.base && mask == a.mask; }
+        constexpr auto operator<=>(const ipv4_net& a) const { return std::tie(base, mask) <=> std::tie(a.base, a.mask); }
 
-        constexpr bool operator!=(const ipv4_net& a) const { return not(*this == a); }
-
-        constexpr bool operator<(const ipv4_net& a) const { return base.to_base(mask) < a.base.to_base(a.mask); }
+        constexpr bool operator==(const ipv4_net& a) const { return std::tie(base, mask) == std::tie(a.base, a.mask); }
 
         constexpr bool contains(const ipv4& addr) const { return addr.to_base(mask) == base; }
     };
@@ -72,17 +73,17 @@ namespace oxen::quic
         {}
 
       public:
-        // Network order
+        // Host order
         uint64_t hi, lo;
 
-        // Host order string constructor
-        ipv6(const std::string& ip);
+        ipv6() = default;
 
-        // Network order in6_addr constructor
+        // Network order in6_addr constructor (calls private constructor)
         ipv6(const struct in6_addr* addr) : ipv6{addr->s6_addr} {}
 
-        constexpr ipv6(
-                uint16_t a = 0x0000,
+        // Explicit to prevent collision with default constructor and make calling purposeful
+        explicit constexpr ipv6(
+                uint16_t a,
                 uint16_t b = 0x0000,
                 uint16_t c = 0x0000,
                 uint16_t d = 0x0000,
@@ -98,11 +99,9 @@ namespace oxen::quic
 
         const std::string to_string() const;
 
-        constexpr bool operator==(const ipv6& a) const { return hi == a.hi && lo == a.lo; }
+        constexpr auto operator<=>(const ipv6& a) const { return std::tie(hi, lo) <=> std::tie(a.hi, a.lo); }
 
-        constexpr bool operator!=(const ipv6& a) const { return not(*this == a); }
-
-        constexpr bool operator<(const ipv6& a) const { return std::tie(hi, lo) < std::tie(a.hi, a.lo); }
+        constexpr bool operator==(const ipv6& a) const { return (*this <=> a) == 0; }
 
         constexpr ipv6 to_base(uint8_t mask) const
         {
@@ -126,18 +125,20 @@ namespace oxen::quic
         ipv6 base;
         uint8_t mask;
 
+        ipv6 max_ip();
+
+        constexpr ipv6_net(ipv6 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}
+
         const std::string to_string() const { return base.to_base(mask).to_string(); }
 
+        constexpr auto operator<=>(const ipv6_net& a) { return std::tie(base, mask) <=> std::tie(a.base, a.mask); }
+
         constexpr bool operator==(const ipv6_net& a) { return base == a.base && mask == a.mask; }
-
-        constexpr bool operator!=(const ipv6_net& a) { return not(*this == a); }
-
-        constexpr bool operator<(const ipv6_net& a) { return base.to_base(mask) < a.base.to_base(a.mask); }
 
         constexpr bool contains(const ipv6& addr) const { return addr.to_base(mask) == base; }
     };
 
-    inline constexpr ipv6_net operator/(const ipv6 a, uint8_t mask)
+    inline constexpr ipv6_net operator/(const ipv6& a, uint8_t mask)
     {
         return {a.to_base(mask), mask};
     }
--- a/include/oxen/quic/ip.hpp
+++ b/include/oxen/quic/ip.hpp
@@ -7,27 +7,34 @@
 
 namespace oxen::quic
 {
-    inline constexpr size_t IPV4_ARRLEN{4};
-    inline constexpr size_t IPV6_ARRLEN{16};
-
     struct ipv4
     {
         // host order
         uint32_t addr;
 
-        ipv4() = default;
+        constexpr ipv4() = default;
 
         constexpr ipv4(uint32_t a) : addr{a} {}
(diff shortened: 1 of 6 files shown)


include/oxen/quic/ip.hpp
```diff
@@ -46,13 +47,15 @@ namespace oxen::quic
         ipv4 base;
         uint8_t mask;
 
+        ipv4 max_ip();
+
         const std::string to_string() const { return base.to_base(mask).to_string(); }
 
-        constexpr bool operator==(const ipv4_net& a) const { return base == a.base && mask == a.mask; }
+        constexpr auto operator<=>(const ipv4_net& a) const { return base.to_base(mask) <=> a.base.to_base(a.mask); }
```
jagerman: There's a bit of weirdness here (not added here, though enlarged on by switching from `<` to `<=>`): by only comparing the base address we get weird things like:

auto a = ipv4{10,0,0,0} / 8;
auto b = ipv4{10,1,0,0} / 16;
auto c = ipv4{10,0,0,0} / 16;

a < b; // true
b < a; // false
a < c; // false;
c < a; // false

and yet most of the addresses in the first range are greater than all of the addresses in the second range, so it feels a little weird to say that "a comes before b".

The `a` <-> `c` comparisons are a bigger problem (not just added here, but in the existing code) for std::map or std::set, because those define an element to exist based on whether `!(a < b) && !(b < a)`.  And so if you try to put `a` and `c` in the same map or set, it will fail because, according to this logic, they are equivalent elements.  `std::sort` would have similar obstacles.

I think you *could* implement <=> with [partial ordering](https://en.cppreference.com/w/cpp/utility/compare/partial_ordering): a comes before b if its entire range comes before b's entire range, and likewise for b.  It's not entirely clear to me what that does for `std::map`/`std::set`/`std::sort`, though, but seems like it would be able to encapsulate the relationship here between ranges.  (I'm not sure we actually have a use case for such a thing, but it's an interesting use case for C++20's partial ordering).

src/ip.cpp
```diff
@@ -67,4 +48,22 @@ namespace oxen::quic
...
+
+        if (mask > 64)
+        {
+            b.hi = base.hi;
+            b.lo |= (128 - mask) - 1;
+        }
+        else
+        {
+            b.hi |= (1 << (64 - mask)) - 1;
+            b.lo = (uint64_t)(int64_t)-1;
```
jagerman: This double cast looks rather weird; I'd suggest instead one of:
```suggestion
            b.lo = std::numeric_limits<uint64_t>::max();
```

```suggestion
            b.lo = static_cast<uint64_t>(-1);
```

src/ip.cpp
```diff
@@ -67,4 +48,22 @@ namespace oxen::quic
...
     }
 
+    ipv6 ipv6_net::max_ip()
+    {
+        auto b = base.to_base(mask);
+
+        if (mask > 64)
+        {
+            b.hi = base.hi;
+            b.lo |= (128 - mask) - 1;
```
jagerman: This is missing a bit shift

src/ip.cpp
dr7ana: My mistake, I should have clarified more in my last discord message. This was my last "stab" at the CI fail -- was trying something from SO. It originally did not have the `int64_t`

include/oxen/quic/ip.hpp
dr7ana: Yeah it's tricky. Some thoughts:

I see some potential weirdness with ordering based on `a comes before b if its entire range comes before b's entire range, and likewise for b`. I think this is a result of trying to assert equality comparisons between sets that may contain one another. As a result, we're using equality comparison 

Using the example ranges from your comment, the range of possible values of `a`  (10.0.0.0 - 10.255.255.255) encompasses that of `b` (10.1.0.0 - 10.1.255.255), so it would be more accurate to say that `b` is a perfect subset of `a`. 

In that consideration, I see your very valid point about the `a` vs `c` comparison and I think we can address it as follows:

```c++
// constructor ensuring ipv4_net always calls ::to_base(...) on construction outside of the `operator/`
constexpr ipv4_net(ipv4 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}

// modified <=> operator
constexpr auto operator<=>(const ipv4_net& a) const 
{
    return std::tie(base, mask) <=> std::tie(a.base, a.mask);
}
```

Now, `a` and `c` produce the same base IP, but their masks are different; therefore `a` would still come prior to `c` by satisfying the statement `a < c == true`.

Thoughts?

jagerman: Why do we need inequality operators at all?

include/oxen/quic/ip.hpp
```diff
@@ -7,36 +7,45 @@
...
         {}
 
+        constexpr std::optional<ipv4> next_ip() const
+        {
+            std::optional<ipv4> ret = std::nullopt;
+
+            if (not will_overflow(addr))
+                ret = ipv4{addr + 1};
+
+            return ret;
```
jagerman: This could be a bit simpler if you just ditched the `will_overflow` and just detect overflow:
```suggestion
            auto next_addr = addr + 1;
            if (next_addr == 0) // overflow
                return std::nullopt;
            return ipv4{next_addr};
```

include/oxen/quic/ip.hpp
```diff
@@ -72,24 +91,33 @@ namespace oxen::quic
...
+            if (not will_overflow(lo))
+                return ipv6{hi, lo + 1};
+
+            // If lo is INT_MAX, then:
+            //  - if hi can be incremented, ++hi and set lo to all 0's
+            //  - else, return nullopt
+            if (not will_overflow(hi))
+                return ipv6{hi + 1, uint64_t{0}};
+
+            return std::nullopt;
```
jagerman: Can also simplify here by dropping `will_overflow` here and let it overflow with detection:
```suggestion
            auto next_lo = lo + 1;
            auto next_hi = hi + (next_lo == 0); // lo overflows into hi
            if (next_hi == 0) // overflow
                return std::nullopt;
            return ipv6{next_hi, next_lo};
```

include/oxen/quic/ip.hpp
```diff
@@ -46,13 +55,21 @@ namespace oxen::quic
...
         uint8_t mask;
 
-        const std::string to_string() const { return base.to_base(mask).to_string(); }
+        constexpr ipv4 max_ip() const
+        {
+            auto b = base.to_base(mask);
 
-        constexpr bool operator==(const ipv4_net& a) const { return base == a.base && mask == a.mask; }
+            if (mask < 32)
+                b.addr |= (1 << (32 - mask)) - 1;
```
```suggestion
                b.addr |= (uint32_t{1} << (32 - mask)) - 1;
```
to match the explicit type in the ipv6 version.  I think this is implicitly okay as is (because `int << uint8_t` gets promoted to `unsigned int << unsigned int`), but the explicit uint32_t there makes it more obviously correct.

src/ip.cpp
```diff
@@ -1,49 +1,20 @@
...
-        if (rv < 0)
-            throw std::invalid_argument{"System error (ec:{}): {}"_format(errno, std::system_category().message(errno))};
-
-        if (rv == 0)
-            throw std::invalid_argument{"IPv4 constructor failed to parse input: {}"_format(ip)};
-    }
-
     const std::string ipv4::to_string() const
     {
+        uint32_t net = oxenc::load_host_to_big<uint32_t>(&addr);
```
jagerman:
```suggestion
        uint32_t net = oxenc::host_to_big(addr);
```
(No need for the heavier memcpy version of the function when you already have a properly aligned integer.)

include/oxen/quic/ip.hpp
```diff
@@ -7,36 +7,45 @@
...
+            return ret;
+        }
+
         const std::string to_string() const;
 
         explicit operator in_addr() const
         {
             in_addr a;
-            a.s_addr = addr;
+            a.s_addr = oxenc::load_host_to_big<uint32_t>(&addr);
```
```suggestion
            a.s_addr = oxenc::host_to_big(addr);
```

include/oxen/quic/ip.hpp
```diff
@@ -46,13 +55,21 @@ namespace oxen::quic
...
+            if (mask < 32)
+                b.addr |= (uint32_t{1} << (32 - mask)) - 1;
 
-        constexpr bool operator!=(const ipv4_net& a) const { return not(*this == a); }
+            return b;
+        }
+
+        constexpr ipv4_net(ipv4 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}
+
+        const std::string to_string() const { return base.to_base(mask).to_string(); }
```
jagerman: - I don't think the `to_base(mask)` here is wanted, since it gets done already in the constructor.
- The mask should be included in the `to_string()` output, e.g. `"10.0.0.0/8"` or `"2001:db8::/32"`

include/oxen/quic/ip.hpp
```diff
@@ -164,20 +204,20 @@ namespace oxen::quic
...
+            ipv6(0x64, 0xff9b, 1, 0, 0, 0, 0, 0) / 48,   // IPv4/IPv6 translation
+            ipv6(0x100, 0, 0, 0, 0, 0, 0, 0) / 64,       // Discard
+            ipv6(0x200, 0, 0, 0, 0, 0, 0, 0) / 7,        // Deprecated NSPA-mapped IPv6; Yggdrasil
+            ipv6(0x2001, 0x0, 0, 0, 0, 0, 0, 0) / 32,    // Toredo
+            ipv6(0x2001, 0x20, 0, 0, 0, 0, 0, 0) / 28,   // ORCHIDv2
+            ipv6(0x2001, 0xdb8, 0, 0, 0, 0, 0, 0) / 32,  // Documentation/example
+            ipv6(0x2002, 0, 0, 0, 0, 0, 0, 0) / 16,      // Deprecated 6to4 addressing scheme
+            ipv6(0xfc00, 0, 0, 0, 0, 0, 0, 0) / 7,       // Unique local address
+            ipv6(0xfe80, 0, 0, 0, 0, 0, 0, 0) / 10,      // link-local unicast addressing
+            ipv6(0xff00, 0, 0, 0, 0, 0, 0, 0) / 8,       // Multicast
```
jagerman: I'm not a big fan of needing to add all these zeros, because it goes against the spirit of being able to leave off trailing 0s in IPv6 addresses (`e.g. 2001:db8::` <-> `ipv6{0x2001, 0xdb8}`).  Perhaps we could make the two-uint64_t constructor take a `std::pair<uint64_t, uint64_t> hi_lo` instead, so that we can avoid the constructor ambiguity with the `uint16_t = 0` arguments?

jagerman: (I also though about making a constructor that would let you do: `ipv6{0x2001, 0xdb8, zeros}` where `zeros` can be specified once anywhere in the argument list with up to eight `uint16_t`s, but couldn't get it to work quickly)

include/oxen/quic/utils.hpp
```diff
@@ -202,6 +202,12 @@ namespace oxen::quic
 
     std::string str_tolower(std::string s);
 
+    template <std::integral T>
+    constexpr bool increment_will_overflow(T val)
+    {
+        return std::numeric_limits<T>::max() - val <= 1;
```
jagerman:
```suggestion
        return std::numeric_limits<T>::max() == val;
```

src/address.cpp
```diff
@@ -68,27 +68,27 @@ namespace oxen::quic
...
 
-    Address::Address(ipv4 v4, uint16_t port)
+    Address::Address(const ipv4& v4, uint16_t port)
     {
         _sock_addr.ss_family = AF_INET;
 
         auto& sin = reinterpret_cast<sockaddr_in&>(_sock_addr);
         sin.sin_port = oxenc::host_to_big(port);
-        std::memcpy(&sin.sin_addr, &v4, sizeof(ipv4));
+        oxenc::write_host_as_big<uint32_t>(v4.addr, &sin.sin_addr);
```
jagerman:
```suggestion
        sin.sin_addr.s_addr = oxenc::host_to_big(v4.addr);
```

src/ip.cpp
```diff
@@ -1,49 +1,20 @@
 #include "ip.hpp"
 
+#include <bit>
```
jagerman: I don't think this is used

include/oxen/quic/ip.hpp
```diff
@@ -7,36 +7,45 @@
 
 namespace oxen::quic
 {
+    inline constexpr size_t IPV4_ARRLEN{4};
+    inline constexpr size_t IPV6_ARRLEN{16};
```
jagerman: Seems unused

include/oxen/quic/utils.hpp
dr7ana: This function is explicitly meant to check if it will overflow after the increment. Re-naming it to fit the purpose was something you brought up earlier, and this change muddies the water in that

include/oxen/quic/ip.hpp
dr7ana: Requiring that the zeros be specified makes the `ipv6` and `ipv4` constructors match each other. Any other swap makes them non-uniform, which was a specific effort planned for this PR in terms of things like host vs net order storage, etc

jagerman (approved):

PR #133, opened 2024-05-29
ipv{4,6} string constructors; const ipv6_range operator==

dr7ana:


src/ip.cpp
```diff
@@ -28,6 +43,22 @@ namespace oxen::quic
         return ret;
     }
 
+    ipv6::ipv6(const std::string& str)
+    {
+        struct sockaddr_in6 sin6
+        {};
```
jagerman: The `struct` keyword isn't wanted, but also this seems like it only needs a `in6_addr` rather than a full `sockaddr_in6`.

src/ip.cpp
```diff
@@ -4,6 +4,21 @@
 
 namespace oxen::quic
 {
+    ipv4::ipv4(const std::string& str)
+    {
+        struct sockaddr_in sin
+        {};
```
jagerman: It doesn't seem like you need a `sockaddr_in` here, but rather just an `in_addr`.

Also the `struct` keyword is not needed/wanted in C++ (which is why clang-format is formatting it incorrectly).

src/ip.cpp
```diff
@@ -4,6 +4,21 @@
...
+        sin.sin_family = AF_INET;
+        auto rv = inet_pton(AF_INET, str.c_str(), &sin.sin_addr);
+
+        if (rv == 0)  // inet_pton returns this on invalid input
+            throw std::invalid_argument{"Cannot construct address: invalid IP"};
+        if (rv < 0)
+            throw std::system_error{errno, std::system_category()};
```
jagerman: This is duplicating some code (mostly the error-to-exception translation) from Address (and also duplicating some code across the v4/v6 versions) that would be nicer to DRY out.  I suggest we extract this into something like this:

```c++
// Wrapper around inet_pton that throws an exception on error
void parse_addr(int af, void* dest, const std::string& from) {
    auto rv = inet_pton(af, from.c_str(), dest);
    // ... check rv and throw if bad
}
// Parses an IPv4 address from string
void parse_addr(in_addr& into, const std::string& from) {
    parse_addr(AF_INET, &into.sin_addr, from);
}
// Parses an IPv6 address from string
void parse_addr(in6_addr& into, const std::string& from) {
    parse_addr(AF_INET6, &into.sin6_addr);
}
```
and then this code becomes just
```c++
    in_addr sin_addr;
    parse_addr(sin_addr, str);
    addr = oxenc::big_to_host(sin_addr.s_addr);
```
and similarly for the `ipv6` and `Address` string constructors.

include/oxen/quic/ip.hpp
```diff
@@ -14,6 +14,8 @@ namespace oxen::quic
 
         constexpr ipv4() = default;
 
+        ipv4(const std::string& str);
```
```suggestion
        explicit ipv4(const std::string& str);
```

include/oxen/quic/ip.hpp
```diff
@@ -93,17 +95,19 @@ namespace oxen::quic
 
         constexpr ipv6() = default;
 
+        ipv6(const std::string& str);
```
```suggestion
        explicit ipv6(const std::string& str);
```

src/ip.cpp
```diff
@@ -4,6 +4,21 @@
...
 {
+    ipv4::ipv4(const std::string& str)
+    {
+        struct sockaddr_in sin
+        {};
+        sin.sin_family = AF_INET;
+        auto rv = inet_pton(AF_INET, str.c_str(), &sin.sin_addr);
+
+        if (rv == 0)  // inet_pton returns this on invalid input
+            throw std::invalid_argument{"Cannot construct address: invalid IP"};
```
jagerman: This message seems off since this isn't constructing an address: perhaps just `"Unable to parse IP address"`

include/oxen/quic/ip.hpp
```diff
@@ -93,17 +95,19 @@ namespace oxen::quic
...
+
         constexpr std::optional<ipv6> next_ip() const
         {
             // If lo will not overflow, increment and return
             if (not increment_will_overflow(lo))
-                return ipv6{{hi, lo + 1}};
+                return ipv6{std::make_pair(hi, lo + 1)};
```
jagerman: I had a thought upon seeing this: we could drop the std::pair constructor if we changed this to:
```
            if (not increment_will_overflow(lo)) {
                ipv6 next{*this};
                next.lo++;
                return next;
            }
```
and similarly for the other one that changes `hi`.

(Purely a suggested alternative: feel free to ignore if you prefer the pair approach).

src/ip.cpp
dr7ana: great suggestion

include/oxen/quic/ip.hpp
dr7ana: i am with you on nuking the pair ctor. Let me change its usage in the test code

jagerman (approved):

PR #134, opened 2024-05-30
Fix crash when stream buffer outlives the sent_req

jagerman:
Because of the timeout on bt request streams, it's possible for the `sent_request` object that owns the request data to get destroyed before the data is acked; subsequent attempts to send will then be referencing invalid memory because the `send()` was only using a view with no ownership.

This transfers the data into the `send()` call so that the stream buffers keep it alive as long as needed, and it doesn't seem like keeping `data` around in the sent_req is needed for anything.

(Found in Session iOS, many thanks to @mpretty-cyro for tracking this down/reporting it)

Edit: also fixes a couple other crashes:
- `when_writable` could fire after the connection is no longer valid.
- `Connection` dropping in Endpoint could happen during Connection methods and needs to be deferred to the event loop.

### DIFF
--- a/include/oxen/quic/btstream.hpp
+++ b/include/oxen/quic/btstream.hpp
@@ -154,9 +154,6 @@ namespace oxen::quic
 
         message to_timeout() && { return {return_sender, ""_bs, true}; }
 
-        std::string_view view() { return {data}; }
-        std::string payload() && { return std::move(data); }
-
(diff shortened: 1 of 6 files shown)


tewinget (approved):

tewinget:
LGTM, not sure what CI is angry about.

PR #136, opened 2024-06-04
GCC stringop bug

dr7ana:
- FMT will trigger a gcc bug on stringop-overflow with gcc versions 11 and earlier
- FMT issue https://github.com/fmtlib/fmt/pull/2442
- FMT issue https://github.com/fmtlib/fmt/issues/2708
- GCC bug tracker https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101854
- Confirming it still appears on gcc-11 https://godbolt.org/z/d15q4Exvz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,8 +112,13 @@ if (WARNINGS_AS_ERRORS)
 endif()
 add_library(libquic_internal-warnings INTERFACE)
 
-if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0.0)
-    target_compile_options(libquic_internal-warnings INTERFACE -fconcepts)
+# gcc 11 and earlier bug
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 12.0.0)
+    if (libquic_IS_TOPLEVEL_PROJECT)
+        list(APPEND warning_flags -Wno-error=stringop-overflow)
+    else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -112,8 +112,13 @@ if (WARNINGS_AS_ERRORS)
...
 add_library(libquic_internal-warnings INTERFACE)
 
-if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0.0)
-    target_compile_options(libquic_internal-warnings INTERFACE -fconcepts)
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0.0)
+        target_compile_options(libquic_internal-warnings INTERFACE -fconcepts)
+    endif()
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 12.0.0)
+        target_compile_options(libquic_external INTERFACE -Wno-error=stringop-overflow) # gcc 11 and earlier bug
```
jagerman: This should get appended to `warning_flags`, like the code just above this is doing for `-Wno-error=array-bounds`, so that it gets the same only-for-c-and-c++ generator expression condition as the other warning flags we set.

dr7ana: It doesn't percolate up to things using libquic as a dependency when it's added to `warning_flags`

I did this the way the libsession example you shared to me earlier was done:
https://github.com/oxen-io/libsession-util/blob/0751f37e82f695ec0212821ad938bbfd96b1e9f5/src/CMakeLists.txt#L10-L17

dr7ana: I gated it with checking if libquic is top-level instead

PR #137, opened 2024-06-05
loop event repeater

dr7ana:
- Repeating events can be instantiated using `::call_every(...)`
- `EventHandler` object is added for lifetime management of repeated events. Application can start/pause/stop at will
- Application has the choice of tying the lifetime of the event to the calling object, or using the `EventHandler`
- 002 test case added sending via repeater
- `std::move` -> `std::forward` where necessary

### DIFF
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -23,20 +23,42 @@ namespace oxen::quic
     static void setup_libevent_logging();
 
     using loop_time = std::chrono::microseconds;
-    using event_ptr = oxen::quic::event_ptr;
     using loop_ptr = std::shared_ptr<::event_base>;
 
-    struct EventHandler
+    class Loop;
+
+    struct EventHandler : public std::enable_shared_from_this<EventHandler>
     {
+        friend class Loop;
+
+      private:
+        std::atomic<bool> _is_running{false};
+        bool _is_stopped{false};
         event_ptr ev;
         timeval interval;
         std::function<void()> f;
 
+        void start_event(
+                const loop_ptr& _loop,
+                loop_time _interval,
+                std::function<void()> task,
+                bool persist = true,
+                bool start_immediately = true);
+
         EventHandler() = default;
 
+      public:
         ~EventHandler();
 
-        void start(const loop_ptr& _loop, loop_time _interval, std::function<void()> task);
+        bool is_running() const { return _is_running and !_is_stopped; }
+
+        bool is_paused() const { return !(_is_running or _is_stopped); }
+
+        bool is_stopped() const { return _is_stopped; }
+
+        bool start();
+        bool pause();
+        bool stop();
     };
 
     class Loop
@@ -52,6 +74,22 @@ namespace oxen::quic
         event_ptr job_waker;
         std::queue<Job> job_queue;
         std::mutex job_queue_mutex;
+
+        template <typename Callable>
+        void add_oneshot_event(loop_time delay, Callable&& hook)
+        {
+            auto handler = make_handler();
+            auto& h = *handler;
+
+            h.start_event(
+                    loop(),
+                    delay,
+                    [hdnlr = std::move(handler), func = std::forward<Callable>(hook)]() mutable {
+                        func();
+                        hdnlr.reset();
+                    },
+                    false);
+        }
 
       public:
         Loop();
@@ -145,6 +183,10 @@ namespace oxen::quic
             return fut.get();
         }
 
+        /** This overload of `call_every` will begin an indefinitely repeating object tied to the lifetime of `caller`.
+            Prior to executing each iteration, the weak_ptr will be checked to ensure the calling object lifetime has
+            persisted up to that point.
+        */
         template <typename Callable>
         void call_every(loop_time interval, std::weak_ptr<void> caller, Callable&& f)
         {
@@ -152,7 +194,8 @@ namespace oxen::quic
             // grab the reference before giving ownership of the repeater to the lambda
             auto& h = *handler;
 
-            h.start(loop(),
+            h.start_event(
+                    loop(),
                     interval,
                     [hndlr = std::move(handler), owner = std::move(caller), func = std::forward<Callable>(f)]() mutable {
                         if (auto ptr = owner.lock())
@@ -162,7 +205,50 @@ namespace oxen::quic
                     });
         }
 
-        void call_soon(std::function<void(void)> f);
+        /** This overload of `call_every` will return an EventHandler object from which the application can start and stop
+            the repeated event. It is NOT tied to the lifetime of the caller via a weak_ptr. If the application wants
+            to defer start until explicitly calling EventHandler::start(), `start_immediately` should take a false boolean.
+        */
+        template <typename Callable>
+        std::shared_ptr<EventHandler> call_every(loop_time interval, Callable&& f, bool start_immediately = true)
+        {
+            auto h = make_handler();
+
+            h->start_event(loop(), interval, std::forward<Callable>(f), true, start_immediately);
+
+            return h;
+        }
+
+        template <typename Callable>
+        void call_later(loop_time delay, Callable&& hook)
+        {
+            if (in_event_loop())
+            {
+                add_oneshot_event(delay, std::forward<Callable>(hook));
+            }
+            else
+            {
+                call_soon([this, func = std::move(hook), target_time = get_timestamp<loop_time>() + delay]() mutable {
+                    auto updated_delay = target_time - get_timestamp<loop_time>();
+
+                    if (updated_delay <= 0us)
+                        func();
+                    else
+                        add_oneshot_event(updated_delay, std::forward<Callable>(func));
+                });
+            }
+        }
+
+        template <typename Callable>
+        void call_soon(Callable&& f)
+        {
+            {
+                std::lock_guard lock{job_queue_mutex};
+                job_queue.emplace(std::forward<Callable>(f));
+            }
+
+            event_active(job_waker.get(), 0, 0);
+        }
 
         void shutdown(bool immediate = false);
 
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -22,6 +22,23 @@ namespace oxen::quic
 
     static void setup_libevent_logging();
 
+    using loop_time = std::chrono::microseconds;
+    using event_ptr = oxen::quic::event_ptr;
+    using loop_ptr = std::shared_ptr<::event_base>;
+
+    struct EventHandler
+    {
+        event_ptr ev;
+        timeval interval;
+        std::function<void()> f;
+
+        EventHandler() = default;
+
+        ~EventHandler();
+
+        void start(const loop_ptr& _loop, loop_time _interval, std::function<void()> task);
+    };
+
     class Loop
     {
         friend class Network;
@@ -46,6 +63,8 @@ namespace oxen::quic
 
         bool in_event_loop() const { return std::this_thread::get_id() == loop_thread_id; }
 
+        std::shared_ptr<EventHandler> make_handler();
+
         // Returns a pointer deleter that defers the actual destruction call to this network
         // object's event loop.
         template <typename T>
@@ -58,7 +77,9 @@ namespace oxen::quic
         template <typename T, typename Callable>
         auto wrapped_deleter(Callable&& f)
         {
-            return [this, f = std::move(f)](T* ptr) { return call_get([f = std::move(f), ptr]() { return f(ptr); }); };
+            return [this, func = std::forward<Callable>(f)](T* ptr) {
+                return call_get([f = std::move(func), ptr]() { return f(ptr); });
+            };
         }
 
         // Similar in concept to std::make_shared<T>, but it creates the shared pointer with a
@@ -77,7 +98,7 @@ namespace oxen::quic
         template <typename T, typename Callable>
         std::shared_ptr<T> shared_ptr(T* obj, Callable&& deleter)
         {
-            return std::shared_ptr<T>(obj, wrapped_deleter<T>(std::move(deleter)));
+            return std::shared_ptr<T>(obj, wrapped_deleter<T>(std::forward<Callable>(deleter)));
         }
 
         template <typename Callable>
@@ -122,6 +143,23 @@ namespace oxen::quic
             });
 
             return fut.get();
+        }
+
+        template <typename Callable>
+        void call_every(loop_time interval, std::weak_ptr<void> caller, Callable&& f)
+        {
+            auto handler = make_handler();
+            // grab the reference before giving ownership of the repeater to the lambda
+            auto& h = *handler;
+
+            h.start(loop(),
+                    interval,
+                    [hndlr = std::move(handler), owner = std::move(caller), func = std::forward<Callable>(f)]() mutable {
+                        if (auto ptr = owner.lock())
+                            func();
+                        else
+                            hndlr.reset();
+                    });
         }
 
         void call_soon(std::function<void(void)> f);
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -8,6 +8,7 @@ extern "C"
 
 #include <atomic>
 #include <cstdint>
+#include <forward_list>
 #include <future>
 #include <memory>
 #include <thread>
@@ -22,42 +23,46 @@ namespace oxen::quic
 
     static void setup_libevent_logging();
 
-    using loop_time = std::chrono::microseconds;
     using loop_ptr = std::shared_ptr<::event_base>;
 
     class Loop;
 
-    struct EventHandler : public std::enable_shared_from_this<EventHandler>
+    struct Ticker
     {
         friend class Loop;
 
       private:
         std::atomic<bool> _is_running{false};
-        bool _is_stopped{false};
         event_ptr ev;
         timeval interval;
         std::function<void()> f;
 
         void start_event(
                 const loop_ptr& _loop,
-                loop_time _interval,
+                std::chrono::microseconds _interval,
                 std::function<void()> task,
                 bool persist = true,
                 bool start_immediately = true);
 
-        EventHandler() = default;
+        Ticker() = default;
 
       public:
-        ~EventHandler();
+        ~Ticker();
 
-        bool is_running() const { return _is_running and !_is_stopped; }
+        bool is_running() const { return _is_running; }
 
-        bool is_paused() const { return !(_is_running or _is_stopped); }
-
-        bool is_stopped() const { return _is_stopped; }
-
+        /** Starts the repeating event on the given interval on Ticker creation
+            Returns:
+                - true: event successfully started
+                - false: event is already running, or failed to start the event
+         */
         bool start();
-        bool pause();
+
+        /** Stops the repeating event managed by Ticker
+            Returns:
+                - true: event successfully stopped
+                - false: event is already stopped, or failed to stop the event
+         */
         bool stop();
     };
 
@@ -75,8 +80,8 @@ namespace oxen::quic
         std::queue<Job> job_queue;
         std::mutex job_queue_mutex;
 
-        template <typename Callable>
-        void add_oneshot_event(loop_time delay, Callable&& hook)
+        template <std::invocable Callable>
+        void add_oneshot_event(std::chrono::microseconds delay, Callable hook)
         {
             auto handler = make_handler();
             auto& h = *handler;
@@ -84,12 +89,18 @@ namespace oxen::quic
             h.start_event(
                     loop(),
                     delay,
-                    [hdnlr = std::move(handler), func = std::forward<Callable>(hook)]() mutable {
+                    [hndlr = std::move(handler), func = std::move(hook)]() mutable {
+                        auto h = std::move(hndlr);
                         func();
-                        hdnlr.reset();
+                        h.reset();
                     },
                     false);
         }
+
+      private:
+        std::list<std::weak_ptr<Ticker>> _tickers;
+
+        void clear_old_tickers();
 
       public:
         Loop();
@@ -101,7 +112,7 @@ namespace oxen::quic
 
         bool in_event_loop() const { return std::this_thread::get_id() == loop_thread_id; }
 
-        std::shared_ptr<EventHandler> make_handler();
+        std::shared_ptr<Ticker> make_handler();
 
         // Returns a pointer deleter that defers the actual destruction call to this network
         // object's event loop.
@@ -112,10 +123,10 @@ namespace oxen::quic
         }
 
         // Returns a pointer deleter that defers invocation of a custom deleter to the event loop
-        template <typename T, typename Callable>
-        auto wrapped_deleter(Callable&& f)
+        template <typename T, std::invocable<T*> Callable>
+        auto wrapped_deleter(Callable f)
         {
-            return [this, func = std::forward<Callable>(f)](T* ptr) {
+            return [this, func = std::move(f)](T* ptr) mutable {
                 return call_get([f = std::move(func), ptr]() { return f(ptr); });
             };
         }
@@ -188,7 +199,7 @@ namespace oxen::quic
             persisted up to that point.
         */
         template <typename Callable>
-        void call_every(loop_time interval, std::weak_ptr<void> caller, Callable&& f)
+        void call_every(std::chrono::microseconds interval, std::weak_ptr<void> caller, Callable&& f)
         {
             auto handler = make_handler();
             // grab the reference before giving ownership of the repeater to the lambda
@@ -210,7 +221,8 @@ namespace oxen::quic
             to defer start until explicitly calling EventHandler::start(), `start_immediately` should take a false boolean.
         */
         template <typename Callable>
-        std::shared_ptr<EventHandler> call_every(loop_time interval, Callable&& f, bool start_immediately = true)
+        [[nodiscard]] std::shared_ptr<Ticker> call_every(
+                std::chrono::microseconds interval, Callable&& f, bool start_immediately = true)
         {
             auto h = make_handler();
 
@@ -219,32 +231,34 @@ namespace oxen::quic
             return h;
         }
 
-        template <typename Callable>
-        void call_later(loop_time delay, Callable&& hook)
+        template <std::invocable Callable>
+        void call_later(std::chrono::microseconds delay, Callable hook)
         {
             if (in_event_loop())
             {
-                add_oneshot_event(delay, std::forward<Callable>(hook));
+                add_oneshot_event(delay, std::move(hook));
             }
             else
             {
-                call_soon([this, func = std::move(hook), target_time = get_timestamp<loop_time>() + delay]() mutable {
-                    auto updated_delay = target_time - get_timestamp<loop_time>();
+                call_soon([this,
+                           func = std::move(hook),
+                           target_time = get_timestamp<std::chrono::microseconds>() + delay]() mutable {
+                    auto updated_delay = target_time - get_timestamp<std::chrono::microseconds>();
 
                     if (updated_delay <= 0us)
                         func();
                     else
-                        add_oneshot_event(updated_delay, std::forward<Callable>(func));
+                        add_oneshot_event(updated_delay, std::move(func));
                 });
             }
         }
 
-        template <typename Callable>
-        void call_soon(Callable&& f)
+        template <std::invocable Callable>
+        void call_soon(Callable f)
         {
             {
                 std::lock_guard lock{job_queue_mutex};
-                job_queue.emplace(std::forward<Callable>(f));
+                job_queue.emplace(std::move(f));
             }
 
             event_active(job_waker.get(), 0, 0);
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -240,15 +240,14 @@ namespace oxen::quic
             }
             else
             {
-                call_soon([this,
-                           func = std::move(hook),
-                           target_time = get_timestamp<std::chrono::microseconds>() + delay]() mutable {
-                    auto updated_delay = target_time - get_timestamp<std::chrono::microseconds>();
+                call_soon([this, func = std::move(hook), target_time = get_time() + delay]() mutable {
+                    auto now = get_time();
 
-                    if (updated_delay <= 0us)
+                    if (now >= target_time)
                         func();
                     else
-                        add_oneshot_event(updated_delay, std::move(func));
+                        add_oneshot_event(
+                                std::chrono::duration_cast<std::chrono::microseconds>(target_time - now), std::move(func));
                 });
             }
         }
--- a/src/loop.cpp
+++ b/src/loop.cpp
@@ -1,7 +1,6 @@
 #include "loop.hpp"
 
 #include "internal.hpp"
-// #include "utils.hpp"
 
 namespace oxen::quic
 {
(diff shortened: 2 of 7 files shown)


include/oxen/quic/loop.hpp
```diff
@@ -124,6 +145,23 @@ namespace oxen::quic
             return fut.get();
         }
 
+        template <typename Callable>
+        void call_every(loop_time interval, std::weak_ptr<void> caller, Callable&& f)
```
jagerman: This interface feels a bit awkward: it seems to be catering to a specific case where I want my timer to be tied to a specific shared_ptr, which means to manage my own timer when it should be tied to something other than a shared_ptr I still get forced into using an external shared ptr unrelated to the timer, so I'd have to do something like:

```C++
struct Dummy {};
auto x = std::make_shared_ptr<Dummy>();
loop->call_every(250ms, x, [] { task(); });

// ...
x.reset();
```
to manage it.  There's also no way I can say "repeat this thing forever" without needing to create and store an unnecessary shared_ptr essentially forever (and even then there's likely a segfault if my shared_ptr outlives the loop), and there's no ability to pause a timer for a while and resume it later.

I do realize this is coming out of Lokinet, but we can do better than simply copy Lokinet's implementation that is catered for Lokinet's specific use case.  (It's actually decently fine for Lokinet's specific use case as is, but a generic library interface needs to be more general).

A possible alternative could work like this:

```C++
// Don't care about controlling it:
loop->call_every(250ms, [] { task(); });

// Option 1: return an ID and add ID-taking control functions to Loop:
auto timer_id = loop->call_every(1s, [] { task(); });
loop->stop_timer(timer_id);
loop->start_timer(timer_id, 500ms);

// Option 2: return an object that manages it:
auto timer = loop->call_every(1s, [] { task(); });
timer->stop();
timer->start(500ms);


// Possibilities for cancelling or pausing a timer from within the timer callback:
class timer_cancel : public std::runtime_error { timer_cancel() : std::runtime_error{"timer cancelled"} {} };
class timer_stop : public std::runtime_error { timer_stop() : std::runtime_error{"timer stopped"} {} };
void task() {
    auto p = some_weak_ptr.lock();
    if (!x)
        throw timer_cancel{};
    //
    // do stuff
    //
    if (++num_calls == 300) {
        num_calls = 0;
        throw timer_stop{};
    }
}
```

src/loop.cpp
```diff
@@ -29,6 +29,44 @@ namespace oxen::quic
...
+        interval = loop_time_to_timeval(_interval);
+
+        ev.reset(event_new(
+                _loop.get(),
+                -1,
+                EV_PERSIST,
+                [](evutil_socket_t, short, void* s) {
+                    auto* self = reinterpret_cast<EventHandler*>(s);
+                    // execute callback
+                    self->f();
```
jagerman: We're in a C callback here: we cannot allow exceptions to bubble up through here.

include/oxen/quic/loop.hpp
```diff
@@ -36,6 +53,18 @@ namespace oxen::quic
...
+        template <typename Callable>
+        void add_oneshot_event(loop_time delay, Callable&& hook)
+        {
+            auto handler = make_handler();
+            auto& h = *handler;
+
+            h.start(loop(), delay, [hdnlr = std::move(handler), func = std::forward<Callable>(hook)]() mutable {
+                func();
+                hdnlr.reset();
+            });
```
jagerman: This feels unnecessarily complex when libevent has the ability to schedule a delayed, non-repeating event.

include/oxen/quic/loop.hpp
dr7ana: I see your point, but creating (option 1) an interface by which loop events are manipulated using their uuids that also need to be implemented or (option 2) an entire timer object is way beyond the scope of what this is meant to effect. 

Tying the timer to the lifetime of an object also makes perfect sense when considering that the logic being repeatedly invoked is also tied to it's lifetime. The test case I added to `002` makes this clear, as the timer repeatedly calls on the stream to send a message

include/oxen/quic/loop.hpp
dr7ana: valid. I added a boolean check to either set the `EV_PERSIST` flag or no flag, which will schedule a one-time non-repeating event

include/oxen/quic/loop.hpp
jagerman: > creating (option 1) an interface by which loop events are manipulated using their uuids that also need to be implemented or (option 2) an entire timer object is way beyond the scope of what this is meant to effect.

Then drop this PR and keep this logic lokinet side.  If we put repeatable events into libquic's event loop then it should be a well-designed interface, and not just copied from the one that Lokinet had.  This implementation in Lokinet was designed when the underlying event code was designed to be swappable to different underlying event loops with potentially different capabilities, which is partly why it has this shared_ptr lifetime management.  It was also good enough for what lokinet needed, but never intended to be a generic event loop timer interface.

> Tying the timer to the lifetime of an object also makes perfect sense when considering that the logic being repeatedly invoked is also tied to it's lifetime. The test case I added to 002 makes this clear, as the timer repeatedly calls on the stream to send a message

Yes, sometimes it does, and the 002 test case is tailored to that use case.  But that's not the only use case for a timer, and a generic library interface has a higher bar in being usable across different use cases.

include/oxen/quic/loop.hpp
```diff
@@ -22,6 +22,45 @@ namespace oxen::quic
 
     static void setup_libevent_logging();
 
+    using loop_time = std::chrono::microseconds;
```
jagerman: Ditch this typedef: the original is a more descriptive type to have in the exposed API than the typedef.

include/oxen/quic/loop.hpp
```diff
@@ -22,6 +22,45 @@ namespace oxen::quic
...
+
+        bool is_running() const { return _is_running and !_is_stopped; }
+
+        bool is_paused() const { return !(_is_running or _is_stopped); }
+
+        bool is_stopped() const { return _is_stopped; }
+
+        bool start();
+        bool pause();
+        bool stop();
```
jagerman, line 66: `stop()` seems unnecessary: its implementation is effectively the same as the destructor, which means an object on which you have called `stop()` is a sort of unusable zombie object.  It would be preferable to use the RAII pattern and have the destructor be the (only) permanent stop.  Then `pause()` can get renamed to `stop()` (which is a nicer linguistic inverse of `start()`), and you simplify the state tracking to just one atomic bool.

include/oxen/quic/loop.hpp
```diff
@@ -36,6 +75,22 @@ namespace oxen::quic
         std::queue<Job> job_queue;
         std::mutex job_queue_mutex;
 
+        template <typename Callable>
+        void add_oneshot_event(loop_time delay, Callable&& hook)
```
jagerman: There's little point here in taking `hook` as a forwarding reference, because you are always capturing it here in the lambda, and so the `Callable&&` followed by the `func = std::forward<Callable>(hook)` capture just delays but doesn't eliminate the need to move or copy -- you still have to move or copy when you construct the lambda.  Thus you can simplify it all by copying or moving at the call site by making this take a non-forwarding reference (and you can also use a concept to make sure its callable with 0 arguments):

```C++
template <std::invocable Callback>
void f(Callback cb) {
    [cb=std::move(cb)] { ... }
}
```

include/oxen/quic/loop.hpp
```diff
@@ -36,6 +75,22 @@ namespace oxen::quic
...
+        {
+            auto handler = make_handler();
+            auto& h = *handler;
+
+            h.start_event(
+                    loop(),
+                    delay,
+                    [hdnlr = std::move(handler), func = std::forward<Callable>(hook)]() mutable {
+                        func();
+                        hdnlr.reset();
```
jagerman: This circular ownership introduces some difficulties:
1. if `func()` throws this leaks because `.reset()` won't get called.
2. if the event loop stops, this leaks.

1. is relatively easy to solve: just steal the shared_ptr before the call the func():
```C++
auto h = std::move(hdnlr);
func();
```
which will always release the shared pointer regardless of how this leaves the lambda.

2. is more difficult.  One solution would be for the Loop object to keep a `std::forward_list<std::weak_ptr<EventHandler>>`, and then:
- just after the loop stops, you walk the list and assign nullptr to the `.f` of any objects that `lock()` successfully.
- each time someone constructs a new EventHandler, you walk the list and remove any weak pointers that have `.expired()` (to keep this from growing indefinitely).

Another approach is what oxenmq does where the EventHandler only lives in the loop and you get some sort of atomic ID to reference it.  (But I think I prefer the weak pointer approach).

include/oxen/quic/loop.hpp
```diff
@@ -58,7 +115,9 @@ namespace oxen::quic
         template <typename T, typename Callable>
         auto wrapped_deleter(Callable&& f)
         {
-            return [this, f = std::move(f)](T* ptr) { return call_get([f = std::move(f), ptr]() { return f(ptr); }); };
+            return [this, func = std::forward<Callable>(f)](T* ptr) {
+                return call_get([f = std::forward<Callable>(func), ptr]() { return f(ptr); });
```
jagerman: See previous comment about the *outer* std::forward here, and about taking a `std::invocable<T*> Callable` by value rather than forwarding reference.

The *inner* std::forward here is actually harmful: if `Callable` is a const lvalue reference then the inner forward here gives you a const value reference to `func`, which forces func to be copied rather than moved into `f`, which definitely isn't what you want.

The outer lambda here should also be `mutable` so that it's `func` can be effectively moved.  (The previous implementation was also copying, because the std::move yields a *const* rvalue reference without `mutable`).

include/oxen/quic/loop.hpp
```diff
@@ -77,7 +136,7 @@ namespace oxen::quic
         template <typename T, typename Callable>
         std::shared_ptr<T> shared_ptr(T* obj, Callable&& deleter)
         {
-            return std::shared_ptr<T>(obj, wrapped_deleter<T>(std::move(deleter)));
+            return std::shared_ptr<T>(obj, wrapped_deleter<T>(std::forward<Callable>(deleter)));
```
jagerman, line 150: This one is good; it's full on forwarding the argument exactly as provided to another function.

include/oxen/quic/loop.hpp
```diff
@@ -22,6 +22,45 @@ namespace oxen::quic
...
+        ~EventHandler();
+
+        bool is_running() const { return _is_running and !_is_stopped; }
+
+        bool is_paused() const { return !(_is_running or _is_stopped); }
+
+        bool is_stopped() const { return _is_stopped; }
+
+        bool start();
+        bool pause();
```
jagerman: These need documenting comments on how they are meant to be used

include/oxen/quic/loop.hpp
```diff
@@ -124,7 +183,75 @@ namespace oxen::quic
...
+                    });
+        }
+
+        /** This overload of `call_every` will return an EventHandler object from which the application can start and stop
+            the repeated event. It is NOT tied to the lifetime of the caller via a weak_ptr. If the application wants
+            to defer start until explicitly calling EventHandler::start(), `start_immediately` should take a false boolean.
+        */
+        template <typename Callable>
+        [[nodiscard]] std::shared_ptr<EventHandler> call_every(
+                loop_time interval, Callable&& f, bool start_immediately = true)
```
jagerman: I think `start_immediately` should be renamed to just `start`, or perhaps inverted to `bool delay_start = false`: the `start_immediately` name made me guess (until I read the code) that this might be doing a sort of "fire now and then repeat every interval".

include/oxen/quic/loop.hpp
```diff
@@ -124,7 +183,75 @@ namespace oxen::quic
...
+            if (in_event_loop())
+            {
+                add_oneshot_event(delay, std::forward<Callable>(hook));
+            }
+            else
+            {
+                call_soon([this,
+                           func = std::forward<Callable>(hook),
+                           target_time = get_timestamp<loop_time>() + delay]() mutable {
+                    auto updated_delay = target_time - get_timestamp<loop_time>();
```
jagerman: `get_time` is the function you want here rather than `get_timestamp` for better type safety.  (`get_timestamp()` is really only there for feeding integral timestamps into ngtcp2 functions).

include/oxen/quic/loop.hpp
```diff
@@ -124,7 +183,75 @@ namespace oxen::quic
...
+
+                    if (updated_delay <= 0us)
+                        func();
+                    else
+                        add_oneshot_event(updated_delay, std::forward<Callable>(func));
```
jagerman: See other comment about a harmful inner forward.

include/oxen/quic/loop.hpp
```diff
@@ -124,7 +183,75 @@ namespace oxen::quic
...
+                });
+            }
+        }
+
+        template <typename Callable>
+        void call_soon(Callable&& f)
+        {
+            {
+                std::lock_guard lock{job_queue_mutex};
+                job_queue.emplace(std::forward<Callable>(f));
```
jagerman: Making this `forward` is counterproductive because it doesn't save anything: because job_queue contains `std::function<void()>`, there's no advantage in forwarding the reference.  All that happens is we carry a reference rather than a value, but then we hit the std::function constructor in the emplace, versus taking this as a `std::function<void()>` in the first place and just moving into the job_queue.

That would be more or less harmless, but because it's being converted from an untemplated function into a generic templated function, that means that pretty much every `call_soon` ends up compiling and either linking or inlining its own unique instantiation of the body of `call_soon` (because every lambda is a different `Callable` type).

src/loop.cpp
```diff
@@ -29,6 +29,111 @@ namespace oxen::quic
...
+            .tv_usec is type __darwin_suseconds_t
+                - this is an equally annoying typedef for `suseconds_t`
+        Alas, yet again another mac idiosyncrasy...
+     */
+    timeval loop_time_to_timeval(loop_time t)
+    {
+        return timeval{
+                .tv_sec = static_cast<decltype(timeval::tv_sec)>(t / 1s),
+                .tv_usec = static_cast<decltype(timeval::tv_usec)>((t % 1s) / 1us)};
+    }
```
jagerman, line 47: I don't understand the comment/code: on both platforms it appears that `tv_sec` is a `time_t` and `tv_usec` is a `suseconds_t` (regardless of the intermediate internal typedef name), so it ought to suffice to `static_cast<time_t>` and `static_cast<suseconds_t>` here.

src/loop.cpp
```diff
@@ -29,6 +29,111 @@ namespace oxen::quic
...
+    bool EventHandler::stop()
+    {
+        if (not _is_running and _is_stopped)
+        {
+            log::critical(log_cat, "EventHandler is already permanently stopped!");
+            return false;
+        }
+
+        ev.reset();
+        f = nullptr;
```
jagerman: This doesn't seem thread-safe: someone could call `stop()` from any thread which means the clearing of the existing `f` could happen while `f` is currently in the middle of a call in the event loop.

(This becomes irrelevant if we get rid of `stop()` in favour of using the destructor, which *is* thread-safe because destruction gets deferred to the event loop).

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+    struct lifetime : public std::enable_shared_from_this<lifetime>
+    {};
+
+    TEST_CASE("013 - EventHandler event repeater: calling object lifetime bound", "[013][repeater][caller]")
+    {
+        Network test_net{};
+        const int NUM_ITERATIONS{10};
+        constexpr auto msg = "hello from the other siiiii-iiiiide"_bsv;
+
+        std::promise<bool> d_promise;
```
jagerman: The `bool` value that goes into this promise never varies and is never read; this should be a `std::promise<void>` instead.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+                ++send_counter;
+                client_stream->send(msg);
+            }
+            else
+                life.reset();
+        });
+
+        test_net.call_later(1s, [&]() {
+            REQUIRE(recv_counter == send_counter);
+            REQUIRE(!life);
```
jagerman: Invoking REQUIRE in another thread like this is a recipe for crashes with catch2; you *can* make it work if you use a mutex to guard anything that touches catch2, but it would be easier to get these values out via a promise and then REQUIRE after the promise gets fulfilled.

tests/013-eventhandler.cpp
jagerman: Actually, given where this is used below, this should stay a `promise<bool>` or maybe a `promise<std::pair<bool, bool>>`, and the conditions currently in the REQUIRE should get propagated via the promise's value to be checked via REQUIRE's outside the call_soon.

tests/013-eventhandler.cpp
jagerman: (And there's a `d_promise` just below this that can do the job)

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+                send_counter += 1;
+                client_stream->send(msg);
+            }
+        });
+
+        REQUIRE(handler->is_running());
+
+        test_net.call_later(1s, [&]() {
+            REQUIRE(recv_counter == send_counter);
+            prom_a.set_value(true);
```
jagerman:
```suggestion
            prom_a.set_value(recv_counter == send_counter);
```

and then use REQUIRE on the value returned from the future.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+        REQUIRE_FALSE(handler->is_running());
+
+        recv_counter = 0;
+        send_counter = 0;
+
+        REQUIRE(handler->start());
+
+        test_net.call_later(1s, [&]() {
+            REQUIRE(recv_counter == send_counter);
+            prom_b.set_value(true);
```
jagerman: As commented above.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+            if (send_counter <= NUM_ITERATIONS)
+            {
+                send_counter += 1;
+                client_stream->send(msg);
+            }
+        });
+
+        REQUIRE(handler->is_running());
+
+        test_net.call_later(1s, [&]() {
```
jagerman: 1s seems too long here, given the 10ms repeat + 10 iterations.  I'd think these test should be able to complete in `2 * 10ms * NUM_ITERATIONS` (200ms; doubled from the ideal 100ms for an error margin), and if they are taking longer than that (but less than 1s) it likely indicates a problem with the event scheduling.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,148 @@
...
+        REQUIRE(handler->is_paused());
+        REQUIRE_FALSE(handler->is_stopped());
+        REQUIRE_FALSE(handler->is_running());
+
+        recv_counter = 0;
+        send_counter = 0;
+
+        REQUIRE(handler->start());
+
+        test_net.call_later(1s, [&]() {
```
jagerman: Too long, see above.

tests/013-eventhandler.cpp
dr7ana: I understand your motivation behind this. Few thoughts on that:
- By the time the `::call_later()` is invoked, the sending and receiving has stopped
- The counters are atomic
- There are literally tons of other non-problematic `REQUIRE` checks in the test code that aren't guarded by mutexes

include/oxen/quic/loop.hpp
dr7ana: Yeah good thinking, thanks for pointing this out

src/loop.cpp
```diff
@@ -29,6 +29,113 @@ namespace oxen::quic
...
+            else
+                ++itr;
+        }
+    }
+
+    std::shared_ptr<Ticker> Loop::make_handler()
+    {
+        clear_old_tickers();
+        auto t = make_shared<Ticker>();
+        _tickers.push_back(t->weak_from_this());
```
jagerman:
```suggestion
        _tickers.push_back(t);
```
But also related, is there a reason `Ticker` needs the added complexity of inheriting from enable_shared_from_this?

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <future>
+#include <oxen/quic.hpp>
+#include <oxen/quic/gnutls_crypto.hpp>
+#include <thread>
+
+#include "utils.hpp"
+
+namespace oxen::quic::test
+{
+    struct lifetime : public std::enable_shared_from_this<lifetime>
```
jagerman: Why enable_shared_from_this?

src/loop.cpp
dr7ana: I explicitly specified this to fix an issue in CI, you can change it back and see how it will fail the mac builds

dr7ana: There are tons of issues online documenting [this](https://github.com/szymonwieloch/rust-rawsock/issues/5)

jagerman: macos seems happy with this:
```bash
om2g:~$ cat timeval.cpp 
```
```C++
#include <sys/time.h>
#include <chrono>
#include <iostream>

using namespace std::literals;

timeval loop_time_to_timeval(std::chrono::microseconds t)
{
    return timeval{
            .tv_sec = static_cast<time_t>(t / 1s),
            .tv_usec = static_cast<suseconds_t>((t % 1s) / 1us)};
}

int main() {
    auto x = loop_time_to_timeval(123456ms);
    std::cout << x.tv_sec << "s, " << x.tv_usec << "µs\n";
}
```
```bash
om2g:~$ g++ -std=c++20 timeval.cpp -o timeval
om2g:~$ ./timeval 
123s, 456000µs
```

Do you have a CI job number where this was failing?

tests/013-eventhandler.cpp
dr7ana: See above response to the `REQUIRE` comment

src/loop.cpp
dr7ana: It's not a lot of overhead, and it simplifies. I don't really see how this is more complex, could you explain?

tests/013-eventhandler.cpp
dr7ana: Above comment

include/oxen/quic/loop.hpp
dr7ana: This feels nit-picky? It's pretty clear what it does, and the comment block explains it explicitly above

src/loop.cpp
jagerman: What does it simplify?

The complexity is in design: rather than being a simple class without inheritance, it's an inheriting class adding a capability that isn't needed or used anywhere.

include/oxen/quic/loop.hpp
dr7ana: I'm using a different `get_timestamp` that takes a template for explicit type safety. Also, the only difference between `get_time` and `get_timestamp` is `.time_since_epoch()`
```c++
std::chrono::steady_clock::time_point get_time()
{
    return std::chrono::steady_clock::now();
}
std::chrono::nanoseconds get_timestamp()
{
    return std::chrono::steady_clock::now().time_since_epoch();
}
```

jagerman: > Also, the only difference between get_time and get_timestamp is .time_since_epoch()

That *is* the important difference.  The latter is an abitrary duration that is meaningless on its own; the former is a object that is a duration with an epoch that fully defines it.

dr7ana: Okay so `get_time` is an arbitrary duration that is meaningless on its own? So makes sense why we shouldn't use that one

dr7ana: I think if you write this out you will see that it's effectively the same

jagerman: No, other way around.  I had former/latter backwards in my original comment.

dr7ana: If you can write this out inline and show me it's simpler and an improvement I will happily make the change, but I don't see your argument on this

jagerman: > I think if you write this out you will see that it's effectively the same

Effectively the same in its final value, but different in its semantic meaning.

You can write all sorts of things that are effectively the same, but the use of types (i.e. abstractions) is to give semantic meaning through your intermediate types of what it is that you hold in your hand.  std::chrono::nanosecond is, under the hood, just an integer, but the type adds meaning to that integer.

A std::chrono::time_point is, under the hood, just a duration with an implied reference point, but that still matters, because nanoseconds with a reference point has a different semantic meaning than nanoseconds alone.

dr7ana: Except that taking an `std::chrono` type (milliseconds, nanoseconds, etc) and duration casting it is the simpler way to do it. Taking a time point requires either appending `.time_since_epoch()` (which makes it *exactly* like `get_timestamp()`) or finding the difference between it and another time_point, at which point one must duration cast...

src/loop.cpp
dr7ana: Can you explain why that is an improvement over the current cast to whatever the system defined `timeval` struct is? The above seems insufficient for other system implementations that can [compile things away](https://codebrowser.dev/gcc/include/x86_64-linux-gnu/bits/types/struct_timeval.h.html), like 64-bit `__time64_t`

jagerman: Because `timeval` is a [documented, decades old POSIX standard](https://pubs.opengroup.org/onlinepubs/7908799/xsh/systime.h.html) that is required to have a `time_t tv_sec;` and `suseconds_t tv_usec;`.  Whatever `__time64_t` or `__darwin_suseconds_t` get used by standard library implementations is irrelevant: those are internal implementation types which are required to be the same as `time_t` and `suseconds_t` typedef in order to satify the `timeval` specification.

dr7ana: Except `time_t` is *itself* a [typedef](https://en.wikipedia.org/w/index.php?title=Time_t&oldid=450752800):

>This type is a [typedef](https://en.wikipedia.org/wiki/Typedef) defined in the standard [<time.h>](https://en.wikipedia.org/wiki/Time.h) header. [ISO C](https://en.wikipedia.org/wiki/ISO_C) defines time_t as an arithmetic type, but does not specify any particular [type](https://en.wikipedia.org/wiki/Data_type), [range](https://en.wikipedia.org/wiki/Range_(mathematics)), [resolution](https://en.wikipedia.org/wiki/Frequency), or encoding for it. Also unspecified are the meanings of arithmetic operations applied to time values.

Casting to the native timeval types is the straight-line, most efficient solution to this.

jagerman: > Except time_t is itself a typedef

Yup.  It's a typedef that is defined to be the type that goes into a `tv_sec`, and `suseconds_t` is likewise a typedef that is defined to be the type that goes into `tv_usec`.

dr7ana: That is also incorrect. Please feel free to [verify this yourself](https://en.cppreference.com/w/c/chrono/time_t). For extra thoroughness, please see the latest [C17 standard](https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf), specifically section 7.27.1-4 to see:

> The range and precision of times representable in clock_t and time_t are implementation-defined

`time_t` is not typedef'ed specifically for `tv_sec`. It is used in a whole host of other functions that can be enumerated by searching "time_t" on the C17 standard pdf file.

On my system, `time_t` is a typedef for `__time_t`, which *itself* typedefs a few times all the way down to `long int` on my system

dr7ana: That is also incorrect, it is not typedef'ed specifically for that use-case. Please feel free to [verify this yourself](https://en.cppreference.com/w/c/chrono/time_t)
> Although not defined by the C standard, this is almost always an integral value holding the number of seconds

For reference to what they're speaking about, please see the latest [C17 standard](https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf), specifically section 7.27.1-4 to see:

> The range and precision of times representable in clock_t and time_t are implementation-defined

`time_t` is not typedef'ed specifically for `tv_sec`. It is used in a whole host of other functions that can be enumerated by searching "time_t" on the C17 standard pdf file.

On my system, `time_t` is a typedef for `__time_t`, which *itself* typedefs a few times all the way down to `long int` on my system

jagerman: Not resolved.

Again: `timeval` is explicitly defined in terms of `time_t` and `suseconds_t`, and thus *those are the types* that belong here.  Not `__time_t`, not `decltype(...)`, not `long int`, not `__darwin_time_t`, but simply the member types that the timeval API explicitly defines, which are `time_t` and `suseconds_t`.

tests/013-eventhandler.cpp
jagerman: Not resolved: if this `bool` value is not being used, this should be a `promise<void>`.

tests/013-eventhandler.cpp
jagerman: Not resolved.

`REQUIRE` inside a callback is something we should avoid if easy to do so precisely because of the Catch2 thread hazard; even if this call does not hit the problem, it is also extremely simple to move the `REQUIRE`s after the `require_future` without sacrificing anything in terms of the test or clarity, but gaining that it is immediately obviously safe since there's no longer any catch2 interactions happening on different threads in the test.

tests/013-eventhandler.cpp
jagerman: Not resolved.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+            if (send_counter <= NUM_ITERATIONS)
+            {
+                ++send_counter;
+                client_stream->send(msg);
+            }
+            else
+                life.reset();
+        });
+
+        test_net.call_later(1s, [&]() {
```
jagerman: Too long: this should match the duration in the other test.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+    TEST_CASE("013 - EventHandler event repeater: EventHandler managed lifetime", "[013][repeater][managed]")
+    {
+        Network test_net{};
+        const int NUM_ITERATIONS{10};
+        const auto INTERVAL{10ms};
+        const auto DELAY{2 * NUM_ITERATIONS * INTERVAL};
+        constexpr auto msg = "hello from the other siiiii-iiiiide"_bsv;
+
+        std::promise<bool> prom_a, prom_b;
+        std::future<bool> fut_a = prom_a.get_future(), fut_b = prom_b.get_future();
```
jagerman: Unused/unneeded `bool` here; these should be `promise<void>`/`future<void>`.

tests/013-eventhandler.cpp
```diff
@@ -0,0 +1,144 @@
...
+            {
+                handler->stop();
+                // if (not have_stopped_handler)
+                // {
+                //     have_stopped_handler = true;
+                // }
+                // else
+                // {
+                //     handler->stop();
+                // }
```
jagerman: Dead code?

src/loop.cpp
dr7ana: Well hasn't this certainly been an educational experience. In the interest of not continuing to block this for a third week over named booleans and other nit-picks, it is changed to read as you want it

dr7ana: [Beautiful](https://ci.oxen.rocks/oxen-io/oxen-libquic/1115/23/3)

dr7ana: [Well](https://ci.oxen.rocks/oxen-io/oxen-libquic/1116/23/3), look like we can securely resolve this now.

dr7ana:
@jagerman ready for merge pending any final comments/thoughts/concerns

include/oxen/quic/loop.hpp
jagerman: I don't know why the comment I wrote on this yesterday isn't here, so posting it again:

The principle here is that you don't throw away type safety before you need to without good reason: which means you should keep time points as time point objects, not durations which lose some semantic meaning.

The code here to keep time points as time points is not any more complicated (and only involves one duration cast instead of two), while not throwing away the type safety in intermediate steps:
```C++
                call_soon([this,
                           func = std::move(hook),
                           target_time = get_time() + delay]() mutable {
                    if (auto now = get_time(); now >= target_time)
                        func();
                    else
                        add_oneshot_event(
                                std::chrono::duration_cast<std::chrono::microseconds>(target_time - now),
                                std::move(func));
                });
```

jagerman (approved):

PR #145, opened 2024-08-14
Delegating event loop destruction to primary Network object

dr7ana:
## __**Problem:**__
Previously, Network objects deferred all closing and cleanup logic to the destructor, in accordance with RAII. This is a good thing, as it simplifies object creation/destruction in relation to the current scope in which it is existing. However, this causes some issues with the ambiguous ownership model with respect to the event loop.

If the Networks exist in different owning objects and/or contexts, the closure of one (ex: if stored in a unique_ptr) will result in the destruction of the shared loop object. This can result in undefined behavior, segfaults, and other wild stuff. Libquic also exposes the ability to create an arbitrary number of Networks using either a pre-existing Loop object, or using the Loop accessor in the Network object

The main functional objects in Libquic (i.e. Endpoints, Connections, Streams, etc) are constructed within shared pointers to be given to the application code. These shared pointers are constructed with custom deleters that schedule the invocation on the event loop. When combined with an ambiguous ownership model, the following may happen:
- If Network A makes a loop, and B is created off of it, B will destroy the loop on invocation of its destructor.
- However, when A invokes its destructor, the endpoint shared pointers will try to `::call()` the deleters, causing a segfault
If Network A or B has any objects persisting in the application code, they will compound the issue when they attempt to close, finding the loop closed as well

## __**Updated Ownership Model:**__
Networks can have one of two relationships to event loop that it manages:

- Standard ownership: The event loop's lifetime is entirely managed within the Network object. It is not a
publicly exposed attribute, and cannot be accessed to directly construct other networks "chained" off of the
same event loop. If the application requires multiple networks that share one underlying event loop, invoking
the `::create_linked_network()` method will return a new Network object sharing its own event loop. For the
application code, this has the effect of only destroying the event loop when the last Network sharing it is
destroyed.
- Application ownership: The event loop's lifetime is entirely managed by the application code and managing
context. A standalone Loop object is created and passed by value in each subsequent Network construction. It
is the responsibility of the application code to ensure that the event loop is last to be destroyed. Objects
like endpoints are held in shared_pointers with deleters scheduled on the event loop. The event loop must
persist until all objects held externally (Network, Endpoint, Connection, etc) are destroyed.

### DIFF
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -83,7 +83,7 @@ namespace oxen::quic
         template <std::invocable Callable>
         void add_oneshot_event(std::chrono::microseconds delay, Callable hook)
         {
-            auto handler = make_handler();
+            auto handler = make_handler(Loop::loop_id);
             auto& h = *handler;
 
             h.start_event(
@@ -98,21 +98,27 @@ namespace oxen::quic
         }
 
       private:
-        std::list<std::weak_ptr<Ticker>> _tickers;
+        static constexpr uint16_t loop_id{0};
+
+        std::unordered_map<uint16_t, std::list<std::weak_ptr<Ticker>>> tickers;
 
         void clear_old_tickers();
 
+        std::shared_ptr<Ticker> make_handler(uint16_t _id);
+
       public:
         Loop();
-        Loop(std::shared_ptr<::event_base> loop_ptr, std::thread::id thread_id);
+
+        Loop(const Loop&) = delete;
+        Loop(Loop&&) = delete;
+        Loop& operator=(Loop&&) = delete;
+        Loop& operator=(Loop) = delete;
 
         virtual ~Loop();
 
         const std::shared_ptr<::event_base>& loop() const { return ev_loop; }
 
         bool in_event_loop() const { return std::this_thread::get_id() == loop_thread_id; }
-
-        std::shared_ptr<Ticker> make_handler();
 
         // Returns a pointer deleter that defers the actual destruction call to this network
         // object's event loop.
@@ -201,19 +207,7 @@ namespace oxen::quic
         template <typename Callable>
         void call_every(std::chrono::microseconds interval, std::weak_ptr<void> caller, Callable&& f)
         {
-            auto handler = make_handler();
-            // grab the reference before giving ownership of the repeater to the lambda
-            auto& h = *handler;
-
-            h.start_event(
-                    loop(),
-                    interval,
-                    [hndlr = std::move(handler), owner = std::move(caller), func = std::forward<Callable>(f)]() mutable {
-                        if (auto ptr = owner.lock())
-                            func();
-                        else
-                            hndlr.reset();
-                    });
+            _call_every(interval, std::move(caller), std::forward<Callable>(f), Loop::loop_id);
         }
 
         /** This overload of `call_every` will return an EventHandler object from which the application can start and stop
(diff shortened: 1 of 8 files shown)


jagerman:
I think that the problems here go deeper than Network and needed to be addressed in Loop itself.

It's not clear to me that a primary field tracking whichever Network object first created the loop will fully solve the issue, because code such as this will still manifest it:
```c++
net1 = std::make_shared<Network>();
auto thread_id = net1->call_get([] { return std::this_thread::get_id(); });
net2 = std::make_shared<Network>(net1->loop(), thread_id);
// sometime later:
net1.reset();
// sometime later still:
net2.reset();
```
Using an external `shared_ptr<Loop>` instance can exhibit a different problem, that shutdown() is *never* called:
```c++
loop = std::make_shared<Loop>();
net1 = std::make_shared<Network>(std::move(loop)); // net1 is now the only owner
```
or
```c++
loop = std::make_shared<Loop>();
net1 = std::make_shared<Network>(loop);
// sometime later:
loop.reset();
```

But the above examples hit some parts of Loop ownership and shutdown that are a little weird as well:
- Loop shutdown() and Loop destruction being different.  In other words, why isn't Loop itself RAII?  (I could maybe see that it might want an ability to shutdown() before destruction, but what I don't see is why destruction isn't just a call to `shutdown(true)`).  If they were unified then shutdown would never need to be called from Network at all: the last Network destruction would give up the last `shared_ptr<Loop>`, thereby destroying/shutting it down.

  I'm suspecting that the reason for the difference might be that somewhere under Network (i.e. probably Connection or Endpoint?) we might have something that expects exclusive ownership of Loop by Network and is expecting a full shutdown() on the Loop during ~Network so that scheduled calls can never come back to us with a Network that no longer exists.  But if we're sharing the loop and letting it outlive Network, that *can* happen and is something we'll need to figure out.
- the first example above currently results in *two* independent Loop instances that are sharing ownership of a libevent `event_base` under the hood, but not sharing an instance of `Loop`.  That feels wrong.
- both of the above points together mean that the change in this PR to not call `shutdown` results in any ticker stops not firing in all three examples.
- (Mostly unrelated) There's a WIN32 bug in `~Loop()` that is doing a WSACleanup, but this is only matching a WSAStartup() in the default constructed Loop case but not the shared-`event_base` constructor.

I'm a bit worried that tracking down and solving the edge cases required for fully reliable external shared ownership of the loop (or even worse, just the event_base) is going to be a fairly heavy job, and I'm almost tempted to scrap the support because of the number of issues that might be lurking in the complexity added around this sort of shared ownership.

Which codebases are currently using/relying on this sort of shared loop ownership?

dr7ana:
@jagerman Deeper issue for sure. Picking up where we left this off-thread, there's a way we can simplify the network/event loop relationship.

dr7ana:
@jagerman Deeper issue for sure. Picking up where we left this off-thread, there's a way we can simplify the network/event loop relationship. Here's what I'm thinking:

- Networks can have one of two relationships to event loop that it manages:
    - Standard ownership: The event loop's lifetime is entirely managed within the Network object. It is not a
        publicly exposed attribute, and cannot be accessed to directly construct other networks "chained" off of the
        same event loop. If the application requires multiple networks that share one underlying event loop, invoking
        the `::create_linked_network()` method will return a new Network object sharing its own event loop. For the
        application code, this has the effect of only destroying the event loop when the last Network sharing it is
        destroyed.
    - Application ownership: The event loop's lifetime is entirely managed by the application code and managing
        context. A standalone Loop object is created and passed by value in each subsequent Network construction. It
        is the responsibility of the application code to ensure that the event loop is last to be destroyed. Objects
        like endpoints are held in shared_pointers with deleters scheduled on the event loop. The event loop must
        persist until all objects held externally (Network, Endpoint, Connection, etc) are destroyed.

There's a bit more detail in the implementation of the commit that I'm pushing with this (ex: tickers now indexed to unique Network id's, internal methods to ensure correct indexing, etc)

include/oxen/quic/loop.hpp
```diff
@@ -98,22 +98,28 @@ namespace oxen::quic
         }
 
       private:
-        std::list<std::weak_ptr<Ticker>> _tickers;
+        static constexpr uint16_t loop_id{0};
```
tewinget: Would be better to use some typedef like `eventloop_id` here and in Network to clarify the purpose.

dr7ana: Yep good call

tewinget (approved):

PR #146, opened 2024-08-23
BTReq Stream Methods

dr7ana:
- Currently, only the number of commands/requests awaiting transmission can be accessed by the application
- Added call-safe getter for number of sent bt requests awaiting response

### DIFF
--- a/include/oxen/quic/btstream.hpp
+++ b/include/oxen/quic/btstream.hpp
@@ -249,7 +249,12 @@ namespace oxen::quic
         /// exception if the endpoint in this message should be considered not found.
         void register_generic_handler(std::function<void(message)> request_handler);
 
+        /// Returns the number of requests/commands that are pending outbound transmission. These have
+        /// NOT been sent yet, rather they have been queued by the application
         size_t num_pending() const;
+
+        /// Returns the number of sent requests awaiting response
+        size_t num_awaiting_response() const;
 
       protected:
         void check_timeouts() override;
@@ -280,5 +285,7 @@ namespace oxen::quic
         size_t parse_length(std::string_view req);
 
         size_t num_pending_impl() const { return user_buffers.size(); }
+
+        size_t num_awaiting_response_impl() const { return sent_reqs.size(); }
     };
 }  // namespace oxen::quic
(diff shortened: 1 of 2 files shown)


include/oxen/quic/btstream.hpp
```diff
@@ -280,5 +285,7 @@ namespace oxen::quic
         size_t parse_length(std::string_view req);
 
         size_t num_pending_impl() const { return user_buffers.size(); }
+
+        size_t num_awaiting_response_impl() const { return sent_reqs.size(); }
```
tewinget, line 289: perhaps this should be sent + unsent, in the off chance that one calls this just as a request is queued to be sent.

dr7ana: That's a good idea, but maybe it makes more sense to add to `::num_pending()`? I can see adding it to `::num_awaiting_response()` defeating the purpose of differentiating them in the first place

If we add your suggestion to `::num_pending()`, the application could just go
`bs->num_awaiting_response() - bs->num_pending()` since `Stream::unsent()` is just `user_buffers.size() - unacked_size`. What do you think?

tewinget: So I guess there's three values of interest here: number of sent pending requests, number of unsent pending requests, and their sum.  I think all three could arguably be useful for different reasons.

dr7ana: I can see that, but isn't the third case doable with the first two?
* num_pending = user_buffers.size();
  - this is the number (not total size) of unsent requests (this is also verified in test case 004) 
* num_awaiting_response = sent_reqs.size();
  - this is the number of successfully sent requests pending response or timeout
* The sum of the above two would be identical to summing the results from the invocations, like `bs->num_awaiting_response() + bs->num_pending()`

There are already two IOChannel methods that inherit through for the values in your first comment (the latter is called `::unsent()` as well, but returns the total data size)

tewinget: Yes, *but* by doing two separate calls as the user you make them not atomic so the sum can change between calls.  That said, I don't think that's really a problem in practice; it should be fine as it is.

Edit to add: especially not really a problem because by the time you make a decision based on that status it can have changed too, in theory, unless you've already done "stop sending new requests" on your end, in which case you won't have a race condition anyway.  So...yeah, scratch that, not a problem at all really.

jagerman (approved):

PR #147, opened 2024-08-29
0RTT, Early data, session ticketing, key verification, stateless reset

dr7ana:
Scope:
- 0rtt/early data: Can be enabled with `opt::enable_0rtt_ticketing` at the endpoint level, and will apply for all connections in/out of that endpoint.
- Session ticketing storage mode: Enabled through the same struct by passing the optional verify/put/get callbacks. The default mode is internal storage managed by libquic, but the application can take responsibility by providing all of the callbacks
- Key verification on/off: Can be disabled with `opt::disable_key_verification` at the connection level by passing the struct to calls to `Endpoint::{listen,connect}(...)`
- Stateless reset can be disabled with `opt::disable_stateless_verification` at the endpoint level


include/oxen/quic/endpoint.hpp
```diff
@@ -230,9 +207,11 @@ namespace oxen::quic
         std::vector<ustring> inbound_alpns;
         std::chrono::nanoseconds handshake_timeout{DEFAULT_HANDSHAKE_TIMEOUT};
 
-        std::map<ustring, ustring> anti_replay_db;
-        std::map<ustring, ustring> encoded_transport_params;
-        std::map<ustring, ustring> path_validation_tokens;
+        std::unordered_map<ustring_view, gtls_ticket_ptr> session_tickets;
```
jagerman: The key here being a string_view seems dangerous: who owns the underlying string the view is pointing into and how do we guarantee it stays valid as long as the key exists in `session_tickets`?

jagerman: Additionally: a string view key is also dangerous as a key because if the underlying string *changes* the unordered_map gets broken.

include/oxen/quic/connection.hpp
```diff
@@ -405,7 +416,13 @@ namespace oxen::quic
         std::atomic<bool> _close_quietly{false};
         std::atomic<bool> _is_validated{false};
 
-        ustring remote_pubkey;
+        ustring remote_pubkey{};
+
+        std::set<int64_t> _early_streams;
```
jagerman: AFAICT this is never actually used for lookup/deduplication/ordering, but rather just as a simple container that gets appended to and then iterated through.  I think a vector would suffice for that instead and be marginally simpler?

include/oxen/quic/types.hpp
```diff
@@ -8,6 +8,25 @@
...
+
+        template <typename T>
+        concept quic_address_type = std::derived_from<T, Address>;
+
+        template <typename T>
+        concept raw_sockaddr_type =
+                std::same_as<T, sockaddr> || std::same_as<T, sockaddr_in> || std::same_as<T, sockaddr_in6>;
+
+        template <typename T>
+        concept gtls_datum_type = std::same_as<T, gnutls_datum_t>;
```
jagerman: This concept doesn't seem very useful compared to just directly using `std::same_as` where used:

```c++
template <concepts::gtls_datum_type T>
operator const T*() const { return &x; }
```
versus:
```c++
template <std::same_as<gnutls_datum_t> T>
operator const T*() const { return &x; }
```

include/oxen/quic/types.hpp
```diff
@@ -8,6 +8,25 @@
 
 namespace oxen::quic
 {
+    class Stream;
+    struct Address;
+
+    namespace concepts
```
jagerman: Grouping all of the concepts in this file just because they are concepts feels wrong to me, because the concepts defined here are not generic but rather are specific to types defined elsewhere.  So in other words, I can't actually touch the `stream_derived_type` context defined here without having first included stream.hpp to make `Stream` available, but in that case, why not define it inside stream.hpp?  (This is similar to the STL: `<concepts>` defines highly generic language concepts, with feature-specific ones, such as range-specific concepts, defined alongside the in feature headers, such as `<range>`).

jagerman: I'm not a huge fan of the separate namespace for concepts because a "concept" is just a generic language feature: it feels a bit like making a separate namespace dedicated only to any function that uses templates, or putting all your integers inside an `integers` namespace.

But I can live with it if you feel strongly.

include/oxen/quic/endpoint.hpp
```diff
@@ -381,11 +376,42 @@ namespace oxen::quic
...
+            }
+        }
+
+        template <concepts::quic_address_type T, typename... Opt>
+        static constexpr void check_address_scheme()
+        {
+            if constexpr ((std::is_same_v<opt::disable_key_verification, std::remove_cvref_t<Opt>> || ...))
+            {
+                static_assert(std::is_same_v<T, Address>, "Disabling key verification requires keyless address!");
+            }
```
jagerman: This makes this fail to compile, which seems good:
```C++
    RemoteAddress a = /*...*/;
    ep->connect(a, opt::disable_key_verification{});
```
but shouldn't it also check that if I use a plain Address, I *am* explicitly specifying `opt::disable_key_verification{}`, so that this is not allowed:
```C++
    Address a = /*...*/;
    ep->connect(a);
```
because I'm pretty sure this will fail, because it will not have `disable_key_verification = true` applied to the context.

That also makes me think that the `opt` here isn't strictly needed because you can always infer it based on whether the address is an `Address` or `RemoteAddress`.  That said: perhaps it's still worth requiring the explicit option to actively disable this, given the implications of the option.

include/oxen/quic/endpoint.hpp
jagerman: To clarify: if this *is* safe (which I suspect it is) this deserves a comment explaining why it's safe.

src/gnutls_session.cpp
```diff
@@ -11,40 +11,81 @@ namespace oxen::quic
...
+        if (htype == GNUTLS_HANDSHAKE_NEW_SESSION_TICKET)
         {
-            if (htype == GNUTLS_HANDSHAKE_NEW_SESSION_TICKET)
+            log::debug(log_cat, "Client received new session ticket from server!");
+            auto* conn = get_connection_from_gnutls(session);
+            auto remote_key = conn->remote_key();
+            auto& ep = conn->endpoint();
+            gnutls_datum_t data{}, encoded{};
+
+            if (auto rv = gnutls_session_get_data2(session, &data); rv != 0)
```
jagerman: This is leaking memory: this gnutls function allocates memory that we are responsible for freeing via `gnutls_free`.

src/gnutls_session.cpp
```diff
@@ -11,40 +11,81 @@ namespace oxen::quic
...
             {
-                auto* conn = get_connection_from_gnutls(session);
-                auto& ep = conn->endpoint();
-                (void)ep;
+                log::warning(log_cat, "Failed to query session data: {}", gnutls_strerror(rv));
+                return rv;
             }
 
-            return 0;
+            if (auto rv = gnutls_pem_base64_encode2(SESSION_TICKET_HEADER, &data, &encoded); rv != 0)
```
jagerman: Needs to be freed when done with.

jagerman: I think we should create a freeing RAII wrapper around gnutls_datum_t for cases like this:
```C++
/// RAII-freeing wrapper for use with gnutls allocating functions.
///
/// Usage:
///
///     gtls_datum data;
///    gnutls_allocating_func(&data.datum);
///    // data will get gnutls_free'd when scope is left.
struct gtls_datum {
    gnutls_datum_t datum{};
    ~gtls_datum() { gnutls_free(d.data); }
};
```
Mainly because its simpler than worrying about needing to free at multiple exit points from the function (early return, final return, exception).

src/gnutls_session.cpp
```diff
@@ -157,20 +212,55 @@ namespace oxen::quic
...
                 throw std::runtime_error("ngtcp2_crypto_gnutls_configure_client_session failed");
             }
+
+            if (_0rtt_enabled and _expected_remote_key)
+            {
+                if (auto maybe_ticket = c.endpoint().get_session_ticket(_expected_remote_key.view()))
+                {
+                    gnutls_datum_t d;
+
+                    if (auto rv = gnutls_pem_base64_decode2(SESSION_TICKET_HEADER, &maybe_ticket->_data, &d); rv != 0)
```
jagerman: Data not freed.

include/oxen/quic/opt.hpp
```diff
@@ -202,5 +202,85 @@ namespace oxen::quic
...
+            Note: If one callback is provided, all the other must be as well. Endpoints are bi-directional in libquic, so
+                the sever-only validate hooks and client-only get/put hooks must be provided together.
+         */
+        struct enable_0rtt_ticketing
+        {
+            std::chrono::milliseconds window{DEFAULT_ANTI_REPLAY_WINDOW};
+
+            gtls_db_validate_cb _check = nullptr;
+            gtls_db_get_cb _fetch = nullptr;
+            gtls_db_put_cb _put = nullptr;
```
jagerman: If these are private, they should be inside `private:`; if they aren't, they should not be `_`-prefixed.

include/oxen/quic/gnutls_crypto.hpp
```diff
@@ -259,20 +260,125 @@ namespace oxen::quic
         }
     };
 
+    struct gtls_session_ticket
```
jagerman, line 303: I find this class a bit more complex than it needs to be: it has a sort of dual mode here where it's a copyable, movable object but then when an external application deals with it, it is expected to ignore that and keep it in a unique_ptr.

I took a stab at simplifying it to simply *always* force it to be used via unique_ptr, which allows dropping some of the pointer updating logic; I've pushed the commit to this PR.

That commit also makes `_key` const, documenting its suitability for use as the key for a map (addressing my earlier comment about the safety of the string_view key).

include/oxen/quic/gnutls_crypto.hpp
```diff
@@ -303,6 +407,7 @@ namespace oxen::quic
         gnutls_anti_replay_t anti_replay;
 
         bool is_client;
+        bool _0rtt_enabled;
```
jagerman: These should be `const` to ensure (1) that they never change after construction, but more interestingly, (2) guaranteeing that they get initialized during construction.

include/oxen/quic/endpoint.hpp
jagerman: I think this is perfectly fine now with my added commit: I left a comment specifically about the safety of using the key as a map.

include/oxen/quic/udp.hpp
```diff
@@ -35,23 +35,25 @@ namespace oxen::quic
     {
         Path path;
         ngtcp2_pkt_info pkt_info{};
-        std::variant<bstring_view, bstring> pkt_data;
+        std::vector<std::byte> pkt_data;
```
jagerman, line 38: I'm a little bit worried about the performance impact of this change, because currently *most* of the code through here doesn't need to copy, while the change here ends up making a copy of every packet data.

I played around with speedtest to see if I could measure it, and it's small but noticeable.  I ran each of these tests three times, sending 5GB across a 10Gbps network link, using -HX on both server/client and -g on the client:

Values are:
1. client user time -- cpu time spent within the process doing things
2. client sys time -- cpu time spent in system mode (i.e. interacting with the kernel, network stack, etc.)
3. server user time
4. server sys time
5. reported speedtest speed

## dev branch
```
0m4.577s 0m1.681s 0m3.873s 0m3.155s 707.493MB/s
0m4.757s 0m1.808s 0m3.785s 0m3.049s 728.825MB/s
0m4.555s 0m1.727s 0m3.720s 0m3.156s 722.932MB/s
```

## 0rtt branch
```
0m4.743s 0m1.941s 0m4.352s 0m3.106s 666.143MB/s
0m4.712s 0m1.645s 0m4.694s 0m3.357s 595.069MB/s
0m4.675s 0m1.845s 0m4.410s 0m3.117s 655.770MB/s
```
so it's not a *huge* decline, but it is there in the user cpu times (and lower overall throughput).

If the intention here was just to remove `bstring`/`bstring_view` use then I suggest going to a `variant<vector<byte>, span<const byte>>` so that we can keep avoiding the memory allocation/copying when not needed.

jagerman: (That said: this is my best guess as to what in this PR might cause the increased CPU usage; I didn't try isolating *just* this change from the rest of the 0rtt PR)

src/connection.cpp
```diff
@@ -41,295 +41,306 @@ namespace oxen::quic
...
+        if (flags & NGTCP2_PATH_VALIDATION_FLAG_PREFERRED_ADDR)
+        {
+            set_remote_addr(path->remote);
+            log::debug(log_cat, "Client set new remote ({}) on successful path validation...", _path.remote);
+        }
+        else
+            log::warning(
+                    log_cat,
+                    "Client path validation {}; no address was provided by server...",
+                    success ? "succeeded" : "failed");
```
jagerman: Is this supposed to be a *warning* when it prints "Client path validation succeeded"?

src/connection.cpp
```diff
@@ -1145,7 +1208,7 @@ namespace oxen::quic
 
         if (uint64_t app_err_code = context->stream_open_cb ? context->stream_open_cb(*stream) : 0; app_err_code != 0)
         {
-            log::info(log_cat, "stream_open_callback returned error code {}, closing stream {}", app_err_code, id);
+            log::warning(log_cat, "stream_open_callback returned error code {}, closing stream {}", app_err_code, id);
```
jagerman: I don't think this should be a log::warning, because this is something the callback is supposed to be able to do to close a stream.  (Or rather: to the extent that is is something warnable is up to the application, not libquic).  Warnings, rather, should be for "something went wrong" messages.

include/oxen/quic/opt.hpp
```diff
@@ -202,5 +202,83 @@ namespace oxen::quic
...
+    using gtls_db_put_cb = std::function<void(gtls_ticket_ptr, time_t)>;
+
+    namespace opt
+    {
+        /** 0-RTT ticketing:
+                The application has two choices in managing 0-RTT ticket storage and the server-side anti-replay db. Passing
+            either the default-constructed or expiry window constructed struct will signal to the endpoints that all storage
+            will happen internally.
+                The application can also pass ALL callbacks and take full responsiblity for management. In doing so, the user
+            must still pass an expiry window value, as that is given directly to gnutls. The following details enumerate the
```
jagerman, line 218: Can you elaborate here on what the expiry window does, and how I (as an application dev) should choose it?

include/oxen/quic/stream.hpp
```diff
@@ -137,6 +137,9 @@ namespace oxen::quic
         size_t unsent_impl() const override;
 
       private:
+        // Called if 0-RTT early data was rejected; marks all sent data as unacked
+        void revert_stream();
```
jagerman, line 147: I'm not sure I understand the comment: wouldn't all early data always be unacked in the case where 0rtt is rejected?  Should the comment read "unsent" rather than "unacked"?

include/oxen/quic/udp.hpp
jagerman: Just looking at the numbers again today, and apparently I didn't look closely enough last night: it's primarily the server user time where there's a fairly large difference (about 20% more CPU time), which supports this being a slowdown in packet processing (since the server does way more packet processing in a speedtest run).

include/oxen/quic/utils.hpp
```diff
@@ -112,6 +112,8 @@ namespace oxen::quic
     inline constexpr std::chrono::seconds DEFAULT_HANDSHAKE_TIMEOUT = 10s;
     inline constexpr std::chrono::seconds DEFAULT_IDLE_TIMEOUT = 30s;
 
+    inline constexpr size_t inverse_golden_ratio = sizeof(size_t) >= 8 ? 0x9e37'79b9'7f4a'7c15 : 0x9e37'79b9;
```
jagerman, line 116:
```suggestion
    // quasi-random fixed value used in multi-value hash computations
    inline constexpr size_t inverse_golden_ratio = sizeof(size_t) >= 8 ? 0x9e37'79b9'7f4a'7c15 : 0x9e37'79b9;
```
Previously it was defined just above where being used, but now it's separated quite a bit so deserves a comment.

include/oxen/quic/gnutls_crypto.hpp
```diff
@@ -3,11 +3,13 @@
 #include <oxenc/base64.h>
```
jagerman, line 3: (This comment isn't necessary for this PR) Now that we have three separate cpp files implementing the stuff in this header, I'm thinking we should split the header up into something like `gnutls/crypto.hpp`/`gnutls/session.hpp`/`gnutls/creds.hpp` to match the cpp split.

include/oxen/quic/gnutls_crypto.hpp
```diff
@@ -303,6 +384,7 @@ namespace oxen::quic
         gnutls_anti_replay_t anti_replay;
 
         bool is_client;
```
jagerman: (Cosmetic, optional) This wasn't done in this PR, but I just noticed this member isn't `_`-prefixed for some reason, while all the other ones are.  Might be a nice change to make it match.

include/oxen/quic/iochannel.hpp
```diff
@@ -1,5 +1,4 @@
 #pragma once
-#include <concepts>
```
jagerman, line 2: This should stay as it's used in the header (in the `call_get_accessor` template).  It *does* get it from some other header currently, but I definitely like including headers for things we directly use.  (Have been burned by this before, in epee, where there are tons of missing includes that break in new STL versions or changes to other components).

include/oxen/quic/opt.hpp
```diff
@@ -202,5 +202,83 @@ namespace oxen::quic
...
+
+            enable_0rtt_ticketing() = default;
+
+            explicit enable_0rtt_ticketing(std::chrono::milliseconds w) : window{w} {}
+
+            explicit enable_0rtt_ticketing(
+                    std::chrono::milliseconds w, gtls_db_validate_cb v, gtls_db_get_cb g, gtls_db_put_cb p) :
+                    window{w}, _check{std::move(v)}, _fetch{std::move(g)}, _put{std::move(p)}
+            {
+                if (not(_check and _fetch and _put))
```
jagerman, line 260: Perhaps:
```suggestion
                if (!(_check && _fetch && _put) && (_check || _fetch || _put))
```
so that a caller can decide at runtime whether they want the callbacks and still use this constructor with either 3 set, or 3 unset but not some set/some not.

jagerman: Perhaps:

```suggestion
                if (!(_check && _fetch && _put) && (_check || _fetch || _put))
```
so that a caller can decide at runtime whether they want the callbacks and still use this constructor with either 3 set, or 3 unset but not some set/some not.

include/oxen/quic/opt.hpp
```diff
@@ -202,5 +202,83 @@ namespace oxen::quic
...
+
+            gtls_db_validate_cb _check = nullptr;
+            gtls_db_get_cb _fetch = nullptr;
+            gtls_db_put_cb _put = nullptr;
+
+            enable_0rtt_ticketing() = default;
+
+            explicit enable_0rtt_ticketing(std::chrono::milliseconds w) : window{w} {}
+
+            explicit enable_0rtt_ticketing(
```
jagerman, line 263:
```suggestion
            enable_0rtt_ticketing(
```
`explicit` doesn't do anything on a non-single-argument constructor.

include/oxen/quic/utils.hpp
jagerman: Looking again: I don't understand why this got moved into `utils.hpp` when it's only an implementation detail of `address.hpp`.

include/oxen/quic/endpoint.hpp
```diff
@@ -94,43 +81,16 @@ namespace oxen::quic
...
-                                        outbound_alpns,
-                                        handshake_timeout,
-                                        remote_pk);
-
-                                p.set_value(it_b->second);
-                                return;
-                            }
-                        }
-                    }
+                    _connect(std::move(remote), qcid, next_rid, p);
```
jagerman, line 93: `_connect` has a very strange return mechanism: not just return by reference, but actually return-by-promise, without making use of any of the capabilities of promise beyond return a value.  Much simpler just to have it return the shared pointer and here do: `p.set_value(_connect(std::move(remote), qcid, next_rid));`, keeping the promise an internal implementation detail of just this one function.

include/oxen/quic/utils.hpp
```diff
@@ -260,3 +262,19 @@ namespace oxen::quic
     }
 
 }  // namespace oxen::quic
+
+namespace std
+{
+    template <>
+    struct hash<oxen::quic::ustring_view>
```
jagerman: I don't think we're allowed to define these, because `ustring_view` and `ustring` are just a typedefs for a `std::basic_string(_view)`, which means these is actually adding a std::hash overloads for an stl type, which seems wrong to me.

If we need to use them as unordered_map or _set keys, then we need instead to use the hasher template argument of the map, such as:
```C++
struct ustring_hasher {
        size_t operator()(const ustring_view& sv) const noexcept
        {
            return hash<std::string_view>{}({reinterpret_cast<const char*>(sv.data()), sv.size()});
        }
};

std::unordered_map<ustring_view, int, ustring_hasher> mymap;
```

include/oxen/quic/udp.hpp
jagerman: An alternative approach is, instead of a variant, have *both* vector and span members: when we need to copy, we copy into the vector and set the span to the vector; when we don't need a copy we just don't set the vector at all and set to the span to whatever input we were fed.  And then `data` can just directly use the span rather than needing to visit the two possibilities.

src/internal.hpp
```diff
@@ -24,26 +24,122 @@ namespace oxen::quic
...
+
+        // Parses an IPv6 address from string
+        inline void parse_addr(in6_addr& into, const std::string& from)
+        {
+            parse_addr(AF_INET6, &into, from);
+        }
+    }  // namespace detail
+
+    struct connection_callbacks
     {
```
jagerman, line 54: The struct rename seems fine, but what is the point of moving them into `internal.hpp` when they are entirely local to connection.cpp?

src/connection.cpp
```diff
@@ -41,295 +41,306 @@ namespace oxen::quic
...
+    int connection_callbacks::on_connection_id_status(
+            ngtcp2_conn* /* _conn */,
+            ngtcp2_connection_id_status_type type,
+            uint64_t /* seq */,
+            const ngtcp2_cid* cid,
+            const uint8_t* token,
+            void* user_data)
+    {
+        auto* conn = static_cast<Connection*>(user_data);
+        assert(conn->stateless_reset_enabled());
```
jagerman, line 179: Why must this be true?

jagerman: I see that this callback is only set up when stateless reset is enabled, but what I don't understand is why that is the case.  I.e. why do we not want to call associate/disassociate_cid when when stateless reset is disabled?

src/context.cpp
```diff
@@ -83,4 +83,13 @@ namespace oxen::quic
         log::trace(log_cat, "IO context stored connection closed callback");
         conn_closed_cb = std::move(func);
     }
+
+    void IOContext::handle_ioctx_opt(opt::disable_key_verification)
+    {
+        log::warning(
+                log_cat,
+                "IOContext disabling key verification for {}bound connections!",
```
jagerman, line 91: Since someone has to explicitly pass this option, it's obviously something they want and it seems a bit weird to warn about it.  Maybe `info` instead?

src/endpoint.cpp
```diff
@@ -85,6 +86,58 @@ namespace oxen::quic
...
+                if (auto exp = gnutls_db_check_entry_expire_time(it->second->datum()); current < exp)
+                {
+                    log::debug(log_cat, "Found existing anti-replay ticket for incoming connection; rejecting...");
+                    return GNUTLS_E_DB_ENTRY_EXISTS;
+                }
+
+                log::debug(log_cat, "Found expired anti-replay ticket for incoming connection");
+            }
+
+            session_tickets[std::move(key)] = std::move(ticket);
```
jagerman:
```suggestion
            session_tickets[key] = std::move(ticket);
```
`key` is a string_view here, so moving it doesn't do anything special.

src/endpoint.cpp
```diff
@@ -138,6 +191,69 @@ namespace oxen::quic
...
+                            outbound_ctx,
+                            outbound_alpns,
+                            handshake_timeout);
+
+                    p.set_value(it_b->second);
+                    return;
+                }
+            }
+        }
+    }
```
jagerman, line 235: Minor thing, but these bodies are nearly almost identical: I think they could get collapsed into something like:
```C++
void _connect(Address a, quic_cid qcid, ..., std::optional<ustring> pk = std::nullopt) {
    // all the stuff
}
void _connect(RemoteAddress ra, quic_cid qcid, ... /* no final optional pk argument here */) {
    Address addr{ra};
    return _connect(addr, qcid, /*...*/, std::move(ra).get_remote_key());
}
```
(Or something similar)

src/endpoint.cpp
```diff
@@ -468,38 +602,77 @@ namespace oxen::quic
...
-                "{} associating CID:{} to {}",
-                dir_str,
-                oxenc::to_hex(cid->data, cid->data + cid->datalen),
+                "{} activating new CID:{} with reset token for {}",
+                conn.is_inbound() ? "SERVER" : "CLIENT",
+                qcid,
                 conn.reference_id());
+        associate_cid(qcid, conn);
+
+        // We only hold one reset token per connection at a time!
```
jagerman, line 597: A question (about the protocol, not the code): why is there only one at a time?

tewinget: I would think that it would be one per active connection ID (which is like default 8 per connection?)

src/endpoint.cpp
```diff
@@ -468,38 +602,77 @@ namespace oxen::quic
...
-    void Endpoint::dissociate_cid(const ngtcp2_cid* cid, Connection& conn)
+    void Endpoint::associate_cid(quic_cid qcid, Connection& conn)
     {
-        if (not cid->datalen)
-            return;
+        assert(in_event_loop());
+        log::trace(
+                log_cat, "{} associating CID:{} to {}", conn.is_inbound() ? "SERVER" : "CLIENT", qcid, conn.reference_id());
+
+        conn_lookup.emplace(qcid, conn.reference_id());
```
jagerman, line 630: Is there any possibility of a duplicate here that should be worried about?

src/stream.cpp
```diff
@@ -303,6 +303,14 @@ namespace oxen::quic
         return std::make_pair(std::move(it), offset);
     }
 
+    void Stream::revert_stream()
+    {
+        assert(endpoint.in_event_loop());
+        log::trace(log_cat, "Stream (ID:{}) reverting after early data rejected...", _stream_id);
+        _unacked_size = 0;
```
jagerman, line 310: I left another comment about this, but it seems strange to me to only reset the unacked size, but not the unsent size: doesn't that mean that there will be an unwanted delay before we naturally try resending the data?

jagerman:
I've now reviewed this all and left various comments.  I think this mostly looks good!

include/oxen/quic/udp.hpp
dr7ana: You're exactly right, and thank you for doing the leg work on getting those results. We spoke about this off band; I was aiming to merge the oxenc PR and do the vector/span member. Will fix 👍

include/oxen/quic/endpoint.hpp
dr7ana: Yes thank you. This was intended safe as enforced usage of `gtls_ticket_ptr` in a unique ptr using the constructor, but your commit more properly parametrized that

include/oxen/quic/types.hpp
dr7ana: I'm not married to the idea. It scratched an itch for me organizationally, but I can inline it to make it a bit less gauche

include/oxen/quic/endpoint.hpp
dr7ana: Good call on the check on `RemoteAddress`. As to the opt, I agree that it's worth requiring the explicit option; this way we keep a logical framework in place to modify/add any functionality to our implementation

src/gnutls_session.cpp
dr7ana: Yep! Great call. But I'm going to change the name of the existing `gnutls_key` to be `gtls_key`, so it matches `gtls_datum`. Generally, I think a nomenclature of `gnutls_*` being GNUTLS library objects and ours being `gtls_*` would help keep them discrete mentally

dr7ana: There are more than few `gnutls_datum_t`'s that we use but do not dealloc (ex: `x509_loader`), fixing those as well

include/oxen/quic/stream.hpp
dr7ana: good catch! thank you

include/oxen/quic/utils.hpp
dr7ana: Mentioned this off-band but putting here: bringing this up into application code, similar to the `to_string_formattable` concept

include/oxen/quic/gnutls_crypto.hpp
dr7ana: My mistake, thank you!

include/oxen/quic/iochannel.hpp
dr7ana: done

include/oxen/quic/opt.hpp
dr7ana: Okay good idea but problem. The put/get hooks obviously access whatever storage is being used, so both or neither need to be passed. Unfortunately, the verify hook [ALSO has to check the database](https://www.gnutls.org/manual/html_node/Anti_002dreplay-protection.html), which means it has to be uniform with the get/put hooks

src/internal.hpp
dr7ana: I wanted to aggregate all the callback declarations (including session and tls callbacks) internally, as they're defined regardless in the `*.cpp` files

include/oxen/quic/utils.hpp
dr7ana: Oops maybe i was too blind with span rage to notice, fixed!

src/connection.cpp
dr7ana: This is for the `on_dcid_status` callback, which is exclusively for stateless reset, and does not call associate/dissociate. There is a delineation between activating/deactivating CIDs and associating/dissociating CIDs. All connections, regardless of stateless reset, associate/dissociate CIDs

Passed for all connections:
https://github.com/dr7ana/libquic/blob/3fb037b09568e276e57c252bc2b260eab597c028/src/connection.cpp#L215
https://github.com/dr7ana/libquic/blob/3fb037b09568e276e57c252bc2b260eab597c028/src/connection.cpp#L227

Endpoint methods:
https://github.com/dr7ana/libquic/blob/3fb037b09568e276e57c252bc2b260eab597c028/src/endpoint.cpp#L605-L676

src/endpoint.cpp
dr7ana: Talked about this off-band: My understanding is stateless reset is tied to the connection itself and renewed with every new DCID. That is why the cb defined by ngtcp2 is `on_dcid_status`

https://github.com/ngtcp2/ngtcp2/blob/67047340bc3f0cffd3217f8be649405a3927065c/lib/ngtcp2_cid.h#L93-L96

Whenever a the new connection id hook is invoked by ngtcp2, it copies the cid token to the current cid stored in the connection, overwriting the old one:
https://github.com/ngtcp2/ngtcp2/blob/67047340bc3f0cffd3217f8be649405a3927065c/lib/ngtcp2_conn.c#L7854-L7857

That's why there's a second map (`reset_token_map` and `reset_token_lookup`) to reference all the other quic cid's

src/endpoint.cpp
dr7ana: It should be many quic_cid's to one ConnectionID (default is 8:1), so I would guess not? Apologies if I'm misunderstanding

src/stream.cpp
dr7ana: Chatted about this awkwardness a few months ago I believe: There is no unsent size, stream calculates `::unsent()` using `size() - unacked()`, so setting unacked to 0 accomplished the same thing

include/oxen/quic/endpoint.hpp
dr7ana: Good call, ty

src/endpoint.cpp
tewinget: Presumably there's an astronomically unlikely chance that there could be a collision, but I *think* CIDs are sent encrypted in both directions so a MITM shouldn't be able to *force* a collision, and either side of a connection intentionally making a new connection and re-using a CID is just messing up their own connection, so it should be fine.  At least, if I understand the question correctly.

include/oxen/quic/connection.hpp
dr7ana: Yep, never used for it. The container ensures that, and the monotonic increasing stream ID allows us to emplace_hint at the back of the container. I can make it a vector if you feel it should be one!

dr7ana:
Closed and reopened #153 because github is stupid and broke updating this PR when trying to drop a commit

PR #150, opened 2024-09-17
Bt-message chaining

dr7ana:
- Unit test circular message chaining verifying bt serialization at each stop
- Based on #147


dr7ana:
cherry-picked into #147

PR #152, opened 2024-12-04
Retyping

dr7ana:
- Based on #153 (previously #147)


include/oxen/quic/utils.hpp
```diff
@@ -51,11 +54,13 @@ namespace oxen::quic
...
-    using bstring = std::basic_string<std::byte>;
-    using ustring = std::basic_string<unsigned char>;
-    using bstring_view = std::basic_string_view<std::byte>;
-    using ustring_view = std::basic_string_view<unsigned char>;
-    using stream_buffer = std::deque<std::pair<bstring_view, std::shared_ptr<void>>>;
+    using namespace oxenc;
+
+    using cspan = oxenc::const_span<char>;
+    using uspan = oxenc::const_span<unsigned char>;
+    using bspan = oxenc::const_span<std::byte>;
```
jagerman: The more I see this `const_span` typedef outside oxenc itself, the more I don't like it.

`oxenc::const_span<C>` is just a typedef for `std::span<const C>`, but it doesn't seem like a useful typedef for external use: it's both longer to type than the thing it aliases; and it introduces a level of apparent, but not actual, differentiation that seems to have negative value.  If you're familiar with `std::span` but not oxenc details, then `std::span<const char>` is instantly understood, but `oxenc::const_span<C>` isn't until you look it up and see that it's just a bit of indirection.  And if you *are* familiar with oxenc internals, then you're almost certainly familiar with std::span.  So who benefits from it?

In the origin WIP PR that added `const_span`, it wasn't just a typedef but was its own class, but that didn't survive to merging.  Within oxen-encoding itself, I suppose it has some marginal utility as declaring "this is the span that oxenc uses", but *outside* oxenc I'm having trouble seeing a point.

Is there some higher purpose or future intention that I'm missing here?

dr7ana: That's a valid argument, and IIRC we talked about all the pros/cons of the suboptimal set of choices for approaching this. Between all the options, an inline namespaced typedef was the least intrusive, and still allowed us to declare comparators for it.

> it's both longer to type than the thing it aliases; and it introduces a level of apparent, but not actual, differentiation that seems to have negative value

I think one of the benefits we deliberated on was precisely that ability to implement type-safe code -- operators, comparators, etc -- around it. I remember that you had elaborated on that point, but again this was a few months ago so apologies if my recollection isn't perfect. In terms of the name length, that's why I like `{c,u,b}span` as application typedefs. In terms of the negative value, I suppose I'm struggling to see the explicit negatives here. The concerns you're mentioning are salient and a continuation of a lot of what we talked about, but if I'm missing the specific negatives or what this typedef gets in the way of doing, could you bring me along?

> Within oxen-encoding itself, I suppose it has some marginal utility as declaring "this is the span that oxenc uses", but outside oxenc I'm having trouble seeing a point

I would surmise that by importing `<oxenc/span.h>`, there's the understanding that the operators and comparators will come with it. That would be a specific  The comparators alone have an idiosyncratic implementation, since the whole "it's equal if the pointers are the same" is not what we were wanting.

The nomenclature also indicates the purpose behind its use contextually, which is a less impactful point but I think one worth mentioning nonetheless.

dr7ana: (Tentatively resolving this, per our last conversation off-band about it)

jagerman: > The concerns you're mentioning are salient and a continuation of a lot of what we talked about, but if I'm missing the specific negatives or what this typedef gets in the way of doing, could you bring me along?

Because it's a piece of indirection that isn't adding something useful.  If someone looking at code sees a function that takes a `oxenc::const_span<std::byte>` then they have to go find out what that type does, whereas if you see a `std::span<const std::byte>` you know instantly what it is, because the latter is standard.  Now here you end up going to the typedef, and then you realize that oh, okay, that's just a `std::span<const std::byte>`.  And so you've introduce a little bit of indirection there to someone working with the code.  Now that sort of indirection isn't a huge negative—it's true of most typedefs of this sort—but needs to have some, at least minor, benefit to outweigh the indirection.

For something like `using cspan = std::span<const char>;` the benefit is that you save yourself a bit of typing, particularly if you use that type often, but you haven't defined a distinct type, you're just declared a shorter way to write it.  And sometimes that is worth it, and becomes more worthwhile as the aliased type gets more complex or is subject to change (thick of code doing `using callback_t = std::function<...>` for some lengthy std::function used several times).  But there is always a tradeoff of the convenience the typedef introduces against the indirection it introduces.

In this case we don't seem to have a benefit; all we've said is "hey if you want to write `std::span<const char>` somewhere now you can write that in your code as `oxenc::const_span<char>` instead."  But what's the advantage?  It's not shorter, it's not clearer, it's not subject to change.

> I think one of the benefits we deliberated on was precisely that ability to implement type-safe code -- operators, comparators, etc -- around it. 

But we didn't end up doing that.  Because `const_span` is merely a typedef, operators or functions defined on it are defined on the actual type, which is `std::span<C>` (which is already broader than `const_span` by not requiring that C be const-qualified).

That is, visibility applies to concrete types, not typedefs.  For example this works perfectly well:

```C++
#include <oxenc/span.h>
#include <iostream>

using namespace oxenc::operators;

int main() {
    std::span<const char> s1{"abc"};
    std::span<const char> s2{"def"};
    std::cout << (s1 == s2 ? "equal" : "not equal") << "\n";
}
```

> I would surmise that by importing <oxenc/span.h>, there's the understanding that the operators and comparators will come with it.

I'm not really following this line of thought; span.h is included (indirectly) by including pretty much any public facing header inside oxenc.  It's only when you additionally add a `using namespace oxenc::operators;` that you get usability of the operators (as in the example above), and then you get the operators generically for any std::span<C> with a comparable C, without ever touching the `const_span` typedef.

And so I'm still left asking the question: what benefit does the typedef have?

dr7ana: > And so I'm still left asking the question: what benefit does the typedef have?

I'm not sure why this is a prescient issue on the libquic retyping. In implementing the oxenc span logic, we had a lot of discussions on pros/cons of each approach. I'm not arguing for this specific approach per se, I was under the impression this was what you wanted? I'm not really clear on why this is a useless indirection, since I felt like you brought up good points as to why it is useful in those discussions.

If not, I'm more than happy to change things. However this would likely require another oxenc PR?

dr7ana: Summing up off-band:
- `const_span<C>` being used inside oxenc itself: I don't love it, but I'm okay with it
- `const_span<C>` being  used outside oxenc itself: avoid, because any such use of `oxenc::const_span<C>` is less immediately clear to me than `std::span<const C>` used in the same spot.
Whereas `bspan` instead of `std::span<const std::byte>` is a useful shortening.

PR #154, opened 2024-12-10
IP ranges and IP networks

dr7ana:
- Differentiated between masked IP addresses ("ip ranges") and masked ip networks ("ip nets")

### DIFF
--- a/include/oxen/quic/ip.hpp
+++ b/include/oxen/quic/ip.hpp
@@ -5,6 +5,28 @@
 #include "formattable.hpp"
 #include "utils.hpp"
 
+/** IP Addressing Types:
+    - ipv{4,6} : These types represent raw ipv4/ipv6 addresses. Both hold their respective addresses in host order
+        internally, converting to network order when needed.
+    - ipv{4,6}_net : These types hold an extra uint8_t netmask, such that they can represent IP addresses with a netmask,
+        such as 1.2.3.4/16, 5.6.7.8/8, and so on. The underlying ipv{4,6} address is stored unmasked as a "networked IP
+        address", rather than an "IP network"
+    - ipv{4,6}_range : These types represent the above mentioned "IP network" (and the entire range of IP addresses within)
+        by storing the masked underlying ipv{4,6} address.
+
+    Example:
+        ipv4 v4{"127.8.69.42"};
+        v4.to_string();             ->      "127.8.69.42"
+
+        ipv4_net v4_net{ipv4, 16};
+        v4_net.to_string();       ->      "127.8.69.42/16"
+        v4_net.ip.to_string();    ->      "127.8.69.42"
+
+        ipv4_range v4_range{ipv4, 16};
+        v4_range.to_string();       ->      "127.8.0.0/16"
+        v4_range.base.to_string();  ->      "127.8.0.0"
+*/
+
 namespace oxen::quic
 {
     struct ipv4
@@ -29,9 +51,6 @@ namespace oxen::quic
             return std::nullopt;
         }
 
-        std::string to_string() const;
-        constexpr static bool to_string_formattable = true;
-
         explicit operator in_addr() const
         {
             in_addr a;
@@ -46,39 +65,10 @@ namespace oxen::quic
         constexpr bool operator==(const in_addr& a) const { return (addr <=> a.s_addr) == 0; }
 
         constexpr ipv4 to_base(uint8_t mask) const { return mask < 32 ? ipv4{(addr >> (32 - mask)) << (32 - mask)} : *this; }
-    };
-
-    struct ipv4_net
-    {
-        ipv4 base;
-        uint8_t mask;
-
-        constexpr ipv4 max_ip() const
-        {
-            auto b = base.to_base(mask);
-
-            if (mask < 32)
-                b.addr |= (uint32_t{1} << (32 - mask)) - 1;
-
-            return b;
-        }
-
-        constexpr ipv4_net() = default;
-
-        constexpr ipv4_net(ipv4 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}
 
         std::string to_string() const;
         constexpr static bool to_string_formattable = true;
-
-        constexpr bool operator==(const ipv4_net& a) const { return std::tie(base, mask) == std::tie(a.base, a.mask); }
-
-        constexpr bool contains(const ipv4& addr) const { return addr.to_base(mask) == base; }
     };
-
-    inline constexpr ipv4_net operator/(const ipv4& a, uint8_t mask)
-    {
-        return ipv4_net{a.to_base(mask), mask};
-    }
 
     struct ipv6
     {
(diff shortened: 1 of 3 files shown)


include/oxen/quic/ip.hpp
```diff
@@ -5,6 +5,28 @@
 #include "formattable.hpp"
 #include "utils.hpp"
 
+/** IP Addressing Types:
+    - ipv{4,6} : These types represent raw ipv4/ipv6 addresses. Both hold their respective addresses in host order
+        internally, converting to network order when needed.
+    - ipv{4,6}_range : These types hold an extra uint8_t netmask, such that they can represent IP-ranges such as
```
```suggestion
    - ipv{4,6}_range : These types hold an extra uint8_t netmask, such that they can represent IP addresses with a netmask such as
```

include/oxen/quic/ip.hpp
```diff
@@ -5,6 +5,28 @@
...
+        1.2.3.4/16, 5.6.7.8/8, and so on. The underlying ipv{4,6} address is stored unmasked as a "networked IP address",
+        rather than an "IP network"
+    - ipv{4,6}_net : These types represent the above mentioned "IP network," storing the masked underlying ipv{4,6}
+        address.
+
+    Example:
+        ipv4 v4{"127.8.69.420"};
```
jagerman: This will throw an exception if you actually try it because 420 isn't a valid octet.

include/oxen/quic/ip.hpp
```diff
@@ -160,48 +147,124 @@ namespace oxen::quic
...
 
-        const std::string to_string() const;
-        constexpr static bool to_string_formattable = true;
+    struct ipv4_range : public detail::masked_ipv4
+    {
+        constexpr ipv4_range() = default;
+        constexpr ipv4_range(ipv4 b, uint8_t m) : detail::masked_ipv4{b, m} {}
 
-        constexpr bool operator==(const ipv6_net& a) const { return std::tie(base, mask) == std::tie(a.base, a.mask); }
+        ipv4_net to_net() const { return ip / mask; }
```
jagerman: Perhaps just `network()` instead of `to_net()`?

include/oxen/quic/ip.hpp
dr7ana: fixed!

include/oxen/quic/ip.hpp
```diff
@@ -160,48 +147,124 @@ namespace oxen::quic
...
+    }  //  namespace detail
+
+    struct ipv4_net : public detail::masked_ipv4
+    {
+        constexpr ipv4_net() = default;
+        constexpr ipv4_net(ipv4 b, uint8_t m) : detail::masked_ipv4{b.to_base(m), m} {}
+    };
 
-        constexpr ipv6_net(ipv6 b, uint8_t m) : base{b.to_base(m)}, mask{m} {}
+    inline constexpr ipv4_net operator/(const ipv4& a, uint8_t mask)
```
jagerman: I think that this operator should return a ipv4_range now instead of a _net, as I think it's a bit more natural for ipv4{1,2,3,4} / 24 to mean an IP with a netmask, rather than the 1.2.3.0/24 network range itself.

include/oxen/quic/ip.hpp
```diff
@@ -160,48 +147,124 @@ namespace oxen::quic
...
+    {
+        return {a.to_base(mask), mask};
+    }
 
-        const std::string to_string() const;
-        constexpr static bool to_string_formattable = true;
+    struct ipv4_range : public detail::masked_ipv4
```
jagerman: This name threw me off a bit, because I was assuming it would represent a range of addresses, but it doesn't: it just represents a ipv4 addr with a netmask.

Perhaps rename to `ipv4_mask` instead, which is short and descriptive enough.

include/oxen/quic/ip.hpp
jagerman: I do realize that gives us no nice operator for representing a base network, though.

That's not the end of the world, but perhaps we could use another operator, such as `&`, for it?

ipv4{127.0.0.1} / 8 == ip with mask, CIDR notation 127.0.0.1/8
ipv4{127.0.0.1} & 8 == ipv4 subnet, CIDR notation 127.0.0.0/8

include/oxen/quic/ip.hpp
```diff
@@ -160,48 +147,124 @@ namespace oxen::quic
...
+
+                return b;
+            }
+
+            constexpr auto operator<=>(const masked_ipv4& other) const
+            {
+                return std::tie(ip, mask) <=> std::tie(other.ip, other.mask);
+            }
+
+            constexpr bool operator==(const masked_ipv4& other) const { return (*this <=> other) == 0; }
```
jagerman: I believe all of this can be replaced with just:
```suggestion
            constexpr auto operator<=>(const masked_ipv4&) const = default;
```
because if you have an explicitly defaulted <=> then == and != get implicitly defined in terms of it (unless otherwise specified).

And I think the same can be done for `ipv4`.

jagerman: On second though: having the <=> in the underlying base class introduces some weirdness we don't want (such as making `net < mask` a valid, if non-nonsensical, comparison), so probably better to just delete it from here completely and define separate comparators in the derived classes so that we don't allow that sort of comparison to compile.

include/oxen/quic/ip.hpp
jagerman: Out-of-band discussion: % for ip-with-mask, / for network-addr.

include/oxen/quic/ip.hpp
jagerman: Out-of-band discussion: swap ipv4_range and ipv4_net names completely (so ipv4_net gets repurposed to the net ip-with-mask, and ipv4_range becomes the old ipv4_net behaviour).

include/oxen/quic/ip.hpp
```diff
@@ -160,49 +147,147 @@ namespace oxen::quic
...
+
+          public:
+            constexpr ipv4 max_ip() const
             {
-                b.hi = base.hi;
-                b.lo |= (uint64_t{1} << (128 - mask)) - 1;
+                auto b = ip.to_base(mask);
+
+                if (mask < 32)
+                    b.addr |= (uint32_t{1} << (32 - mask)) - 1;
```
jagerman, line 182: Not added in this PR, but just noticed this in passing:

This should actually be `- 2` instead of `- 1`, because the .255 isn't actually an addressable IP within the range.

It is, rather, the broadcast IP, so perhaps we could do `- 2` here, and then also add one that explicitly calls it out as the broadcast address:

```c++
// Returns the network broadcast address for this subnet, i.e. the largest address in the range:
constexpr ipv4 broadcast() const {
    auto b = ip.to_base(mask);
    if (mask < 32)
        b.addr |= (uint32_t{1} << (32 - mask)) - 1;
    return b;
}
```

jagerman (approved):

include/oxen/quic/ip.hpp
dr7ana: done

PR #158, opened 2025-03-21
De-Network & public Network close()

jagerman:
This fixes #129 but ends up being a rather drastic simplification of various internals to make it work, and in the process made PR #129 less important.

This PR simplifies libquic's API by making the underlying Loop more
accessible in various places, and removing the various loop method
wrappers from Network and Endpoint so that things wanting `call_get`,
for example, now get at it through network.loop().call_get() or
endpoint.loop.call_get() or similar.

After removing those, I noticed that there was very little left in
Network, and so I pushed a little further and made Network entirely
optional: it is now possibly to simply manage some endpoints yourself.
Network is still around, but all it really does now is act as a fancy
container that does a graceful close on endpoints it creates during
destruction.

Some related changes here:

- objects constructed via `loop.make_shared<T>(...)` (previously that or
  net.make_shared or endpoint.make_shared) now destruct *synchronously*,
  while previously they only initiated destruction.  The previous
  behaviour caused issues, though, because an application had no way to
  know when destruction actually happened.  For example, destroying the
  last shared_ptr<Endpoint> and then trying to use the same address
  failed (in the test code) because the destruction of the shared_ptr
  only transferred that shared_ptr off to the event loop but destruction
  would happen later.
- made Endpoint's constructor private, with a new Endpoint::endpoint
  static function to construct one (correctly wrapped in a shared_ptr).
- renamed `loop.in_event_loop()` to `loop.inside()` which works as it is
  now pretty much always accessed through a variable or pointer called
  "loop".
- rename `loop.loop()` to `loop.get_event_base()` to reflect what it
  actually does, and return it as a raw pointer rather than shared_ptr.
  (Also now hold it in a unique_ptr instead of a shared_ptr because a
  Loop should never be sharing that part of itself).
- removed the concept of a "linked network".  If you want two networks
  to share a loop, just create a loop and give it to both, or create a
  network and give its loop to the second one.
- as a consequence of the above, Network no longer tries to micromanage
  its Loop, instead just sharing ownership and using the Loop
  destruction to stop the loop.
- removed the per-network ticker support entirely.  It added a bunch of
  stuff (like "loop ids") that really isn't needed as any application
  can manage its tickers itself and tying them to the network (rather
  than the loop) doesn't add much value on the outside, while adding
  complexity internally.
- removed Network manual shutdown of the loop; the loop now shuts down
  during destruction.
- Removed a dubious promise during loop thread startup that doesn't
  appear to have any effect.
- Remove some do-nothing lambda `mutable`s


tewinget:
The changes look fine, pending conflict resolution & CI

tewinget (approved):

PR #161, opened 2025-03-26
Updated oxen-encoding, patch GnuTLS for NDK 27+, fixes for NDK 26

mpretty-cyro:
With version `27` and above the Android NDK includes definitions for `timezone_t` and a various other time-with-timezone functions but locks everything except `timezone_t` behind an API level 35 check - this results in build errors with GnuTLS when building for Android because it assumes if `timezone_t` is available then the other functions are as well.

To resolve this I've added a `patch` which gets applied for Android builds that forces the `HAVE_TIMESTAMP_T` value to `0` (so that it includes `time_rz.c` in the build) and renames the internal GnuTLS definition of `timezone_t` to `rpl_timezone_t` (to avoid the name collision) which resolves the build issue for NDK 27+.

For NDK 26 the `lexicographical_compare` for `std::vector` doesn't exist so we need to use custom logic in that case to support building for NDK 26.

This PR also updates `oxen-encoding` to the latest version which cleans up some `span` logic (and makes the changes needed to build using the updated version).

### DIFF
--- a/include/oxen/quic/address.hpp
+++ b/include/oxen/quic/address.hpp
@@ -339,36 +339,10 @@ namespace oxen::quic
         RemoteAddress(RemoteAddress&& other) = default;
         RemoteAddress& operator=(RemoteAddress&& other) = default;
 
-        auto operator<=>(const RemoteAddress& other) const
-        {
-#if defined(__ANDROID__) && __NDK_MAJOR__ < 27
-            auto ret = std::strong_ordering::equal;
-
-            if (_remote_pubkey.size() != other._remote_pubkey.size())
-            {
-                ret = _remote_pubkey.size() < other._remote_pubkey.size() ? std::strong_ordering::less
-                                                                          : std::strong_ordering::greater;
-            }
-            else
-            {
-                for (size_t i = 0; i < _remote_pubkey.size(); ++i)
-                {
-                    if (_remote_pubkey[i] != other._remote_pubkey[i])
-                    {
-                        ret = _remote_pubkey[i] < other._remote_pubkey[i] ? std::strong_ordering::less
-                                                                          : std::strong_ordering::greater;
-                        break;
-                    }
-                }
-            }
-#else
-            auto ret = _remote_pubkey <=> other._remote_pubkey;
-#endif
-            if (ret == 0)
-                ret = Address::operator<=>(other);
-            return ret;
+        auto operator<=>(const RemoteAddress& other) const = delete;
+        auto operator==(const RemoteAddress& other) const {
+            return Address::operator==(other) && _remote_pubkey == other._remote_pubkey;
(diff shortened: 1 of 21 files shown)


include/oxen/quic/address.hpp
```diff
@@ -341,7 +341,29 @@ namespace oxen::quic
...
+                for (size_t i = 0; i < _remote_pubkey.size(); ++i)
+                {
+                    if (_remote_pubkey[i] != other._remote_pubkey[i])
+                    {
+                        ret = _remote_pubkey[i] < other._remote_pubkey[i] ? std::strong_ordering::less
+                                                                          : std::strong_ordering::greater;
+                        break;
+                    }
+                }
+            }
```
jagerman: I assume the problem here is that `<=>` isn't working on a vector?

I don't think it makes a lot of sense to have this `<=>` operator at all here, because inequality comparisons on a RemoteAddress don't make a lot of sense to me, *and* because explicitly deleting this... doesn't seem to cause any problems at all, so let's just go that way instead rather than using an Android workaround for a function we don't need or use:

```C++
auto operator<=>(const RemoteAddress& other) const = delete; // Explicit to mask Address's <=>
bool operator==(const RemoteAddress& other) const {
    return Address::operator==(other) && _remote_pubkey == other._remote_pubkey;
}
```

include/oxen/quic/format.hpp
```diff
@@ -44,7 +42,7 @@ namespace oxen::quic
         explicit buffer_printer(std::basic_string<T>&& buf) = delete;
 
         // Constructed from any type of span
-        template <oxenc::const_span_type T>
+        template <oxenc::bt_input_string T>
```
jagerman, line 45: This feels out of place, but I suppose this is only temporary until de-network PR is ready which replaces this (and the later similar changes) entirely.

include/oxen/quic/address.hpp
mpretty-cyro: Yep pretty much - no worries, I'll make the change now then

include/oxen/quic/format.hpp
mpretty-cyro: Yep, pretty much

include/oxen/quic/address.hpp
mpretty-cyro: Ah, looks like `libSession` uses `<=>` at the moment to sort vectors of `service_node` (which inherits from `RemoteAddress`) in order to get a unique set for the snode pool - but I can just add a custom comparator for that function  instead

jagerman: That custom comparator sounds preferable to a sort-of-arbitrary sort order being defined here.

jagerman (approved):

PR #163, opened 2025-04-04
add endpoint opt to disable path mtu discovery

tewinget:


jagerman:
Reports are that using this as a workaround for QUIC connectivity problems, particular under some VPN connections and 4G internet, significantly improves the situation.

I think we should revisit eventually to see if we can do something better than bare minimum, but merely *having* this option available for applications to use as a workaround does not mean it has to be used.

jagerman (approved):

PR #166, opened 2025-07-18
Disable GSO at runtime by default

jagerman:
GSO can fail, depending on the hardware/software: for instance, storage server packets sent via GSO into a Lokinet tun will fail depending on kernel version.

This commit makes GSO usage a runtime endpoint option that has to be opted-in, and is disabled by default.  (And of course opting is does nothing if GSO isn't supported on the system).


jagerman:
"Hide whitespace" highly recommended on the diff.

tewinget (approved):

PR #167, opened 2025-08-29
Make ngtcp2_cid internal to quic_cid

jagerman:
Directly exposing the inheritance is questionable and leads to confusing code.  This uses composition instead of inheritance, and makes the conversion to `const ngtcp2_cid*` explicit where needed via a `ngtcp2()` method.


jagerman:
Included in #168

PR #168, opened 2025-09-04
Lokinet updates

jagerman:
Improvements/changes for Lokinet:
- clean up quic_cid handling
- downgrade a warning that fairly frequently comes up but really isn't something warnable
- fix ALPN not being configurable on a per-connection basis


tewinget (approved):
LGTM

PR #169, opened 2025-09-11
Downgrade some log warnings

jagerman:
These being warnings mean they show up in application logs, and really shouldn't as they are just business-as-usual handling of things that can happen on the network.

### DIFF
--- a/src/endpoint.cpp
+++ b/src/endpoint.cpp
@@ -457,9 +457,12 @@ namespace oxen::quic
 
         if (written <= 0)
         {
-            log::warning(
+            // This error comes up rather frequently under normal operations, as ngtcp2 can decide
+            // that we aren't allowed to send anything right now, so keep it at merely debug log
+            // level.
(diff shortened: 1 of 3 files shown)


src/endpoint.cpp
```diff
@@ -1062,7 +1065,7 @@ namespace oxen::quic
 
         if (not _manual_routing and !socket)
         {
-            log::warning(log_cat, "Cannot sent to dead socket for path {}", p);
+            log::warning(log_cat, "Cannot send to dead socket for path {}", p);
```
tewinget, line 1068: This one still feels kinda warning-y, unless there's some common circumstance under which it could/should be able to happen?

jagerman: That one was just a typo fix, it's still warning.

tewinget: Ah, I'm blind.

tewinget (approved):

PR #171, opened 2025-09-15
Add wakeable event; remove weird/unused rescheduling ticker

jagerman:
Adds quic::Wakeable/loop.make_wakeable() that lets you create an idempotent triggerable event so that you can fire at it many times without worrying about triggering it multiple times.

Also removes the "task_rescheduling" option from Trigger because it isn't used anywhere and was fairly useless.

Also removes Trigger's _is_running bool, because `start()` and `stop()` are underlying calls to libevent that are already idempotent and don't need an atomic bool around them.

### DIFF
--- a/include/oxen/quic/loop.hpp
+++ b/include/oxen/quic/loop.hpp
@@ -52,7 +52,7 @@ namespace oxen::quic
     /// An event loop task that can be fired multiple times, but is triggered manually when needed
     /// to schedule a callback call on the event loop.  Unlike using `call`/`call_soon`, calls to
     /// trigger/wake the event are idempotent: i.e. the event will be called only once regardless of
-    /// how many wakeups there were prior to the call.  Once called, it will not be schedule again
+    /// how many wakeups there were prior to the call.  Once called, it will not be scheduled again
     /// until triggered at least once more.
     ///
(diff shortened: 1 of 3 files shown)


include/oxen/quic/loop.hpp
```diff
@@ -22,41 +21,55 @@ namespace oxen::quic
...
                 - true: event successfully stopped
                 - false: event is already stopped, or failed to stop the event
          */
         bool stop();
     };
 
+    /// An event loop task that can be fired multiple times, but is triggered manually when needed
+    /// to schedule a callback call on the event loop.  Unlike using `call`/`call_soon`, calls to
+    /// trigger/wake the event are idempotent: i.e. the event will be called only once regardless of
+    /// how many wakeups there were prior to the call.  Once called, it will not be schedule again
```
tewinget: will not be schedule(d)

src/loop.cpp
```diff
@@ -215,6 +193,10 @@ namespace oxen::quic
             }
         }
 
+        for (auto* ods : delayed_events)
```
tewinget: Not that it matters, but I presume this was supposed to be "osd" rather than "ods"?

tewinget (approved):

PR #173, opened 2025-09-17
No-credential connections and 0-RTT fixes

jagerman:
This commit adds the ability for a client to connect to a server without providing a key, and allows listening servers to choose whether client keys are required, requested, or omitted.  Required means a client must provide a key, and it must pass the key verification function; requested means the server will request a key and the client may provide one: the key verification function is called with an empty key if the client does not provide one (so that it can still use the ALPN to reject the connection, for example, if certain ALPNs require keys and others don't).  Omitted means the server won't request one at all.

For omitted client pubkeys (either not requested, or requested and not provided but still allowed), the connection's `remote_key()` will return an empty span.

Client connections can now use GNUTLSCreds::make_unauthenticated to make a connection without a pubkey.  They can also omit the TLSCreds from the connect() call entirely, but that won't allow doing things like enable 0rtt.  This connection with such a missing pubkey will fail if the server requires pubkeys, but otherwise is allowed.

This deliberately renames `set_key_verify_callback` to `require_client_keys` (thus breaking code) as calling code needs to decide whether it wants requires or requests behaviour.

This also fixes closely related issues with 0-RTT: under 0-RTT resumption, key verification callbacks were never called, and so the remote_key was never set on the connection object.  This is actually intentional: a valid 0-RTT ticket already contains the connection data, and so the initial connection that produced the ticket already passed verification.  This commit contains fixes to this so that they get properly reflected in the Connection object, and documents this for the key verification functions.

Other changes:

- remove pointless `gtls_key` struct and make it back into a simple typedef, like it used to be.  The indirection added nothing.

- replace Connection `is_validated()` with `is_handshaked()` and `is_handshake_confirmed()`.  Validated isn't used anywhere outside of libquic's test suite, and doesn't really work with 0-RTT connections that don't re-validate.

- remove `set_draining()` and `set_closing()` -- they were pointless private methods that simply set the variable of the same name to false.  But anything with private access can already just set those to false itself.

- Fixed wrong comment above cert_verify_callback_gnutls about what a failure return value looks like.  The function was already returning 1 rather than a negative (which is actually okay), but this intent is now clearer in the comment and the code.

- Removed unused `usage_flags`.  It was already tagged `[[maybe_unused]]` but was actually unused and didn't seem to have any purpose.

### DIFF
--- a/include/oxen/quic/gnutls_crypto.hpp
+++ b/include/oxen/quic/gnutls_crypto.hpp
@@ -133,23 +133,7 @@ namespace oxen::quic
         }
     };
 
-    struct gtls_key final : std::array<unsigned char, GNUTLS_KEY_SIZE>
-    {
-        gtls_key() = default;
-        gtls_key(const unsigned char* data, size_t size) { write(data, size); }
-        explicit gtls_key(std::string_view data) : gtls_key{reinterpret_cast<const unsigned char*>(data.data()), data.size()}
-        {}
-        explicit gtls_key(std::span<const unsigned char> data) : gtls_key{data.data(), data.size()} {}
-
-        //  Writes to the internal buffer holding the gnutls key
-        void write(const unsigned char* buf, size_t size)
-        {
-            if (size != GNUTLS_KEY_SIZE)
-                throw std::invalid_argument{"GNUTLS key must be 32 bytes!"};
-
-            std::memcpy(data(), buf, size);
-        }
-    };
+    using gtls_key = std::array<unsigned char, GNUTLS_KEY_SIZE>;
 
     // key: remote key to verify, alpn: negotiated alpn's
     using key_verify_callback = std::function<bool(std::span<const unsigned char> key, std::string_view alpn)>;
@@ -367,10 +351,17 @@ namespace oxen::quic
         friend class GNUTLSSession;
 
         GNUTLSCreds(std::string_view ed_seed, std::string_view ed_pubkey);
+        GNUTLSCreds();
 
       public:
+        // Construct a credentials object from a Ed25519 seed and pubkey.  (The seed may optionally
+        // be a combined seed+pk libsodium 64-byte value; only the first 32 bytes are used).
         static std::shared_ptr<GNUTLSCreds> make_from_ed_keys(std::string_view seed, std::string_view pubkey);
+        // Construct a credentials object from a Ed25519 combined seed/pubkey value.
         static std::shared_ptr<GNUTLSCreds> make_from_ed_seckey(std::string_view sk);
+        // Constructs a no-credentials object.  This object may *only* be used for client
+        // connections, and will fail if the server is configured to require a client certificate.
+        static std::shared_ptr<GNUTLSCreds> make_unauthenticated();
 
         using anti_replay_add_cb = std::function<bool(
                 std::span<const unsigned char> key,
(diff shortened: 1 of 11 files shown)


include/oxen/quic/gnutls_crypto.hpp
```diff
@@ -367,10 +351,17 @@ namespace oxen::quic
...
+        GNUTLSCreds();
 
       public:
+        // Construct a credentials object from a Ed25519 seed and pubkey.  (The seed may optionally
+        // be a combined seed+pk libsodium 64-byte value; only the first 32 bytes are used).
         static std::shared_ptr<GNUTLSCreds> make_from_ed_keys(std::string_view seed, std::string_view pubkey);
+        // Construct a credentials object from a Ed25519 combined seed/pubkey value.
         static std::shared_ptr<GNUTLSCreds> make_from_ed_seckey(std::string_view sk);
+        // Constructs a no-credentials object.  This object may *only* be used for client
+        // connections, and will fail if the server is configured to require a client certificate.
```
tewinget: s/and will fail/and connections using it will fail/

tewinget (approved):

src/connection.cpp
```diff
@@ -1354,6 +1354,8 @@ namespace oxen::quic
             assert(ins);
             return 0;
         }
+        else if (id == next_incoming_stream_id)
```
tewinget, line 1368: This will work if we only ever allow bidirectional streams.  I wonder if the better answer is to have two queues for incoming streams (uni and bidi) and not assign an id to them at all, but rather if there is a pending incoming stream on the queue when a stream is opened, we pop it, use it, and assign its id at that time.  I don't *think* anything is relying on that queued stream having an id set before it's opened.

src/connection.cpp
```diff
@@ -495,6 +489,19 @@ namespace oxen::quic
         return 0;
     }
 
+    void Connection::check_established()
+    {
+        if (establish_hook_called)
```
tewinget, line 494: both call sites are also doing this check, so it's slightly redundant (but probably good to keep anyway?)

tewinget: nevermind I'm blind.

jagerman: There are a couple others that don't check it, and the ones that do are basically doing it to add a log statement for the "weird" case here (and the tiny gain of not making another function call).

src/connection.cpp
jagerman: Just to leave a comment here about why this is tracking IDs:

The reason for tracking the stream is to deal with ngtcp2 not calling stream opened callbacks for implicitly opened streams.

I.e. in QUIC, if you see new data on stream 8, then the QUIC protocol implies opening of stream 0 and 4.  (+1/2/3 to all of these for unidirectional and/or server-initiated), and ngtcp2 fires 8 but *declines* to fire 0 and 4 stream opened callbacks (although it is marked as subject to change in the future).  If you're trying to match up streams between client and server (i.e. server and client treat stream 0 as a BTReqStream, stream 4 as some custom stream, and stream 8 as a plain stream) then things would break as a result.

Thus tracking the ID, so that if the client opens three streams (which will always be 0-4-8) and the server queues three incoming streams (0-4-8), everything matches up without requiring ngtcp2 to call them in order.

As to unidirectional streams: yeah, this is indeed broken, but currently I don't think we handle unidirection streams at all, and we would need to either add more counters, or else just deprecate this in favour of using stream construction callbacks instead.

PR #175, opened 2025-09-26
add queued datagram bytes limit

tewinget:
The datagram channel backing up is not only a memory usage issue, but also will cause congestion control algorithms for a connection over top of this (e.g. a quic connection over lokinet, which in turn means the datagrams go over multiple other quic connections) to throttle angrily.  Ideally this will help to mitigate that latter issue, but it will probably need tuning.

### DIFF
--- a/src/datagram.cpp
+++ b/src/datagram.cpp
@@ -239,6 +239,10 @@ namespace oxen::quic
             {
                 // Early data accepted, so now we can discard all the sent packets (which will be
                 // everything up, but not including, `early_data_head`).
+                [[maybe_unused]] size_t old_unsent = unsent_bytes;
+                for (auto pkt_itr = buf.begin(); pkt_itr < buf.begin() + *early_data_head; pkt_itr++)
+                    unsent_bytes -= pkt_itr->size();
+                assert(unsent_bytes <= old_unsent);  // in case I'm dumb and the loop above is one too many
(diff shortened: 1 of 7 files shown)


src/datagram.cpp
```diff
@@ -256,13 +264,18 @@ namespace oxen::quic
             {
                 // Early data accepted, so now we can discard all the sent packets (which will be
                 // everything up, but not including, `early_data_head`).
+                [[maybe_unused]] size_t old_unsent = unsent_bytes;
+                for (auto pkt_itr = buf.begin(); pkt_itr < buf.begin() + *early_data_head; pkt_itr++)
+                    unsent_bytes -= pkt_itr->size();
```
jagerman: Slightly nicer:
```suggestion
                for (auto& pkt : buf | std::views::take(*early_data_head))
                    unsent_bytes -= pkt.size();
```

src/datagram.cpp
```diff
@@ -92,6 +94,12 @@ namespace oxen::quic
                 return;
             }
 
+            if (unsent_impl() > dgram_queue_limit)
+            {
+                log::trace(log_cat, "Dropping datagram, queue over limit.");
```
jagerman: I think this should be definitely higher than trace.  debug at least, but maybe even info?

jagerman (approved):

PR #176, opened 2025-09-30
Rewrite and expose size + data accumulation

jagerman:
This is useful for custom streams that have a similar need to decode a prefix-length block of data but aren't actually a BTRequestStream.

This also rewrites the accumulation to be simpler.

Finally this allows (and ignores) an empty BTRequestStream data (e.g. `0:`).  This was a fatal BTRequestStream error before, but could be useful, for instance as a no-op to keep a connection alive without a quic-level keep-alive.

### DIFF
--- a/include/oxen/quic/btstream.hpp
+++ b/include/oxen/quic/btstream.hpp
@@ -34,9 +34,9 @@ namespace oxen::quic
     // request sizes
     inline constexpr long long MAX_REQ_LEN = 10_M;
 
-    // maximum length of the bencoded request length string, including the `:`.  This must be large
-    // enough to hold `MAX_REQ_LEN` followed by a `:`.
-    inline constexpr size_t MAX_REQ_LEN_ENCODED = 9;  // "10000000:"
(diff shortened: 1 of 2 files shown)


include/oxen/quic/btstream.hpp
```diff
@@ -180,6 +180,38 @@ namespace oxen::quic
...
+
+    /// Accumulates data from a stream until reaching the given size.  This is typically used
+    /// immediately after `prefix_accumulator` to accumulate the amount of data indicated by the
+    /// prefix.  If the incoming data does not complete the size, it is appended to `buf` for a
+    /// repeated call to continue to accumulate (or complete).
+    ///
+    /// `req` is reduced by the size of accumulated data (which might, for instance, be the
+    /// beginning of the next request).
+    ///
+    /// Returns true if `buf` has been filled to the requested size, nullopt if more data is needed.
```
tewinget: obviously can't return nullopt, so I assume this should say false if more data needed

tewinget (approved):

